PR #4, opened 2016-07-07
Add PMTCT app

NicolaasDieBaas:
- add flows
- add tests
- add validation/functionality placeholder code


NicolaasDieBaas:
all commits to this points came from a merge with open issue 1 (setup base)

NicolaasDieBaas:
PMTCT flow will need to be adapted if it's been updated and differs from SoW doc

NicolaasDieBaas:
@imsickofmaps @gsvr please review.  You'll see quite a bit of commented blocks of code, these serves as the placeholders for needed functionality to be added as a next step

src/ussd_pmtct.js
```diff
@@ -0,0 +1,254 @@
...
+        App.call(self, "state_start");
+        var $ = self.$;
+        // var interrupt = true;
+
+        self.init = function() {
+            // init services
+        };
+
+    // TEXT CONTENT
+        var questions = {
```
gsvr: @NicolaasDieBaas Do you feel like this setup of having the question text up here and not in the states adds value? A few times with our other apps like this I've thought it's not worth it.

go-app-ussd_pmtct.js
```diff
@@ -0,0 +1,267 @@
...
+                        & utils.check_number_in_range(content, "1900", utils.get_today().year)
+                },*/
+                next: function(content) {
+                    self.im.user.set_answer("dob_year", content);
+                    return "state_birth_month";
+                }
+            });
+        });
+
+        self.add("state_birth_month", function(name) {
```
gsvr, line 154: I notice that the text here doesn't align with the flow diagram - did you get content elsewhere? eg `Mar` instead of `March`.

go-app-ussd_pmtct.js
```diff
@@ -0,0 +1,267 @@
...
+                choices: [
+                    new Choice("yes", $("Yes")),
+                    new Choice("no", $("No"))
+                ],
+                next: function(choice) {
+                    if (choice.value === "yes") {
+                        self.im.user.set_answer("hiv_opt-in", "true");
+                        return "state_end_hiv_messages_confirm";
+                    } else {
+                        self.im.user.set_answer("hiv_opt-in", "false");
```
gsvr: I'm not sure if the dash will cause problems at some point - it probably won't but I like to just stick with underscores.

gsvr:
@NicolaasDieBaas Review done. :+1: when you've made the changes and Travis is happy.

gsvr:
@NicolaasDieBaas There is no timeout state in the scope, so we'll need to limit character counts to 140. In conversation with Praekelt about getting new text.

src/ussd_pmtct.js
NicolaasDieBaas: @gsvr I guess if we're not going to add state error types eventually, then I'm happy to rather have the text in the states themselves

PR #7, opened 2016-07-13
Add PMTCT registration utils and functionality

NicolaasDieBaas:
- [x] assess functionality requirements
  - [x] service api calls
  - [x] util functions

### DIFF
--- a/go-app-ussd_pmtct.js
+++ b/go-app-ussd_pmtct.js
@@ -10,6 +10,7 @@ go.app = function() {
     var EndState = vumigo.states.EndState;
     var FreeText = vumigo.states.FreeText;
     var JsonApi = vumigo.http.api.JsonApi;
+    var HttpApi = vumigo.http.api.HttpApi;
 
     var IdentityStore = require('@praekelt/seed_jsbox_utils').IdentityStore;
     var StageBasedMessaging = require('@praekelt/seed_jsbox_utils').StageBasedMessaging;
@@ -65,7 +66,7 @@ go.app = function() {
             var subscription_base_url = self.im.config.vumi.subscription_url;
             var endpoint = "subscription/";
 
-            var http = new JsonApi(im, {
+            var http = new HttpApi(im, {
                 headers: {
                     'Content-Type': ['application/json'],
                     'Authorization': ['ApiKey ' + username + ':' + api_key]
@@ -150,7 +151,8 @@ go.app = function() {
                             .then(function(has_active_subscription) {
                                 if (has_active_subscription) {
                                     // get details (lang, consent, dob) & set answers
-                                    self.im.user.set_lang(identity.details.lang || "en"); // if undefined default to english
+                                    self.im.user.set_answer("language_choice", identity.details.lang || "en");  // if undefined default to english
+                                    self.im.user.set_lang(self.im.user.answers.language_choice);
                                     self.im.user.set_answer("consent", identity.details.consent !== undefined ? identity.details.consent : false);
                                     self.im.user.set_answer("dob", identity.details.dob !== undefined ? identity.details.dob : null);
 
@@ -177,7 +179,8 @@ go.app = function() {
                                 .then(function(active_subscription_count) {
                                     if (active_subscription_count > 0) {
                                         // save contact data (set_answer's) - lang, consent, dob
-                                        self.im.user.set_lang(contact.data[0].extra.language_choice || "en");
+                                        self.im.user.set_answer("langauge_choice", contact.data[0].extra.language_choice || "en");
+                                        self.im.user.set_lang(self.im.user.answers.language_choice);
                                         self.im.user.set_answer("consent", contact.data[0].consent !== undefined ? contact.data[0].consent : false);
                                         self.im.user.set_answer("dob", contact.data[0].dob !== undefined ? contact.data[0].dob : null);
 
--- a/go-app-ussd_pmtct.js
+++ b/go-app-ussd_pmtct.js
@@ -9,15 +9,82 @@ go.app = function() {
     var PaginatedChoiceState = vumigo.states.PaginatedChoiceState;
     var EndState = vumigo.states.EndState;
     var FreeText = vumigo.states.FreeText;
+    var JsonApi = vumigo.http.api.JsonApi;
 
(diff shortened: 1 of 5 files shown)


NicolaasDieBaas:
Travis checks will fail because 'seed-jsbox-utils' still need to be added to package.json when ready

go-app-ussd_pmtct.js
```diff
@@ -6,17 +6,86 @@ go.app = function() {
...
+        // variables for services
+        var is;
+        var sbm;
 
         self.init = function() {
-            // init services
+            // initialising services
+            var base_url = self.im.config.services.identity_store.prefix;
+            var auth_token = self.im.config.services.identity_store.token;
+            is = new IdentityStore(new JsonApi(self.im, null), auth_token, base_url);
```
gsvr: It may be safer to use `{}` here instead of `null`

src/ussd_pmtct.js
```diff
@@ -3,17 +3,86 @@ go.app = function() {
...
+
+            base_url = self.im.config.services.stage_based_messaging.prefix;
+            auth_token = self.im.config.services.stage_based_messaging.token;
+            sbm = new StageBasedMessaging(new JsonApi(self.im, null), auth_token, base_url);
+        };
+
+
+        // get/load contact from vumigo
+        self.getVumiContactByMsisdn = function(im, msisdn) {
+            var token = "abcde";  // config
```
gsvr: Don't forget to replace this.

src/ussd_pmtct.js
```diff
@@ -3,17 +3,86 @@ go.app = function() {
...
+            var base_url = self.im.config.services.identity_store.prefix;
+            var auth_token = self.im.config.services.identity_store.token;
+            is = new IdentityStore(new JsonApi(self.im, null), auth_token, base_url);
+
+            base_url = self.im.config.services.stage_based_messaging.prefix;
+            auth_token = self.im.config.services.stage_based_messaging.token;
+            sbm = new StageBasedMessaging(new JsonApi(self.im, null), auth_token, base_url);
+        };
+
+
```
gsvr, line 37: Please put a comment in here that indicates that the functions below are essentially temporary while we're still loading information from the old system.

src/ussd_pmtct.js
```diff
@@ -3,17 +3,86 @@ go.app = function() {
...
+        };
+
+        self.getVumiActiveSubscriptionCount = function(im, msisdn) {
+            var username = "superman";
+            var api_key = "";
+
+            var subscription_base_url = "https://foo/";
+            var endpoint = "subscription/";
+
+            var http = new HttpApi(im, {
```
gsvr: Use JsonApi?

src/ussd_pmtct.js
```diff
@@ -3,17 +3,86 @@ go.app = function() {
...
+                }
+            });
+        };
+
+        self.getVumiActiveSubscriptionCount = function(im, msisdn) {
+            var username = "superman";
+            var api_key = "";
+
+            var subscription_base_url = "https://foo/";
+            var endpoint = "subscription/";
```
gsvr, line 63: Replace the above with config values?

src/ussd_pmtct.js
```diff
@@ -57,42 +126,78 @@ go.app = function() {
...
+                .then(function(identity) {
+                    if (identity.details.pmtct) {  // on new system and PMTCT?
+                        // optout
+                        return self.states.create("state_optout_reason_menu");
+                    } else {  // register
+                        return sbm.has_active_subscription(identity.id)
+                            .then(function(active_subscription) {
+                                if (active_subscription) {
+                                    // get details (lang, consent, dob) & set answers
+                                    self.im.user.set_lang(identity.details.lang !== undefined ? identity.details.lang : "en" ); // if undefined default to english
```
gsvr: The shorthand for this is: `self.im.user.set_lang(identity.details.lang || "en")`.

src/ussd_pmtct.js
```diff
@@ -57,42 +126,78 @@ go.app = function() {
...
+            var msisdn = utils.normalize_msisdn(self.im.user.addr, '27');
+            self.im.user.set_answer("msisdn", msisdn);
+
+            return is.get_or_create_identity({"msisdn": msisdn})
+                .then(function(identity) {
+                    if (identity.details.pmtct) {  // on new system and PMTCT?
+                        // optout
+                        return self.states.create("state_optout_reason_menu");
+                    } else {  // register
+                        return sbm.has_active_subscription(identity.id)
```
gsvr, line 145: This will cause us headaches in the future if we're not careful. I'm okay with letting it slide for now for the demo since the only active subscriptions in the new system will be PMTCT subscriptions, but we really should be checking for an active PMTCT subscription, not just any active subscription.  Put a Todo in here and open an issue for that and we'll handle it next week.

src/ussd_pmtct.js
```diff
@@ -57,42 +126,78 @@ go.app = function() {
...
+                            .then(function(active_subscription) {
+                                if (active_subscription) {
```
gsvr: Please change the variable name to `has_active_subscription`

src/ussd_pmtct.js
```diff
@@ -57,42 +126,78 @@ go.app = function() {
...
+                    if (contact.data.length > 0) {
+                        // check if registered on MomConnect
+                        if (contact.data[0].extra.is_registered) {
+
+                            // get subscription to see if active
+                            return self.getVumiActiveSubscriptionCount(self.im, msisdn)
+                                .then(function(active_subscription_count) {
+                                    if (active_subscription_count > 0) {
+                                        // save contact data (set_answer's) - lang, consent, dob
+                                        self.im.user.set_lang(contact.data[0].extra.language_choice !== undefined ? contact.data[0].extra.language_choice : "en");
```
gsvr: Use the shorthand here, too.

NicolaasDieBaas:
@gsvr please review.  You might note that after i reverted back to using HttpApi instead of JsonApi for the getVumiActiveSubscriptionCount function... I was getting a strange error message when using JsonApi and for time's sake switched back for now

gsvr:
:+1:

PR #21, opened 2016-08-03
PMTCT stop saving user account to identity

NicolaasDieBaas:
We should only be saving the contact key - the user account is the essentially MomConnect's account number.


NicolaasDieBaas:
@gsvr please review. i removed the saving of vumi user account to the identity and also fixed the saving of vumi contact key to identity

gsvr:
:+1:

PR #23, opened 2016-08-05
NurseConnect Welcome States

NicolaasDieBaas:
check all status'

### DIFF
--- a/go-app-ussd_nurse.js
+++ b/go-app-ussd_nurse.js
@@ -33,21 +33,21 @@ go.app = function() {
 
             var msisdn = utils.normalize_msisdn(self.im.user.addr, '27');
             return is
-                .get_or_create_identity({"msisdn": msisdn})
-                .then(function(identity) {
-                    if ((!_.isUndefined(identity.nc_working_on))
-                        && (identity.nc_working_on !== "")) {
-                        self.im.user.set_answer("user", identity);
-                        return is
-                            .get_or_create_identity({"msisdn": identity.nc_working_on})
-                            .then(function(working_on_identity) {
-                                self.im.user.set_answer("contact", working_on_identity);
-                            });
-                    } else {
-                        self.im.user.set_answer("user", identity);
-                        self.im.user.set_answer("contact", identity);
-                    }
-                });
+            .get_or_create_identity({"msisdn": msisdn})
+            .then(function(identity) {
+                if ((!_.isUndefined(identity.nc_working_on))
+                    && (identity.nc_working_on !== "")) {
+                    self.im.user.set_answer("user", identity);
+                    return is
+                    .get_or_create_identity({"msisdn": identity.nc_working_on})
+                    .then(function(working_on_identity) {
+                        self.im.user.set_answer("contact", working_on_identity);
+                    });
+                } else {
+                    self.im.user.set_answer("user", identity);
+                    self.im.user.set_answer("contact", identity);
+                }
+            });
         };
 
         // override normal state adding
@@ -68,7 +68,7 @@ go.app = function() {
         self.add('state_route', function(name) {
             // reset working_on extra
             self.im.user.set_answer("nc_working_on", "");
-            
+
             var user = self.im.user.answers.user;
             if (user.nc_is_registered === 'true') {
                 return self.states.create('state_subscribed');
--- a/go-app-ussd_nurse.js
+++ b/go-app-ussd_nurse.js
@@ -17,7 +17,7 @@ go.app = function() {
     var utils = SeedJsboxUtils.utils;
 
     var GoNDOH = App.extend(function(self) {
-        App.call(self, "state_route");
+        App.call(self, "state_start");
         var $ = self.$;
 
         // variables for services
@@ -30,24 +30,6 @@ go.app = function() {
                 self.im.config.services.identity_store.token,
                 self.im.config.services.identity_store.url
             );
-
-            var msisdn = utils.normalize_msisdn(self.im.user.addr, '27');
-            return is
-            .get_or_create_identity({"msisdn": msisdn})
-            .then(function(identity) {
-                if ((!_.isUndefined(identity.nc_working_on))
-                    && (identity.nc_working_on !== "")) {
-                    self.im.user.set_answer("user", identity);
-                    return is
-                    .get_or_create_identity({"msisdn": identity.nc_working_on})
-                    .then(function(working_on_identity) {
-                        self.im.user.set_answer("contact", working_on_identity);
-                    });
-                } else {
-                    self.im.user.set_answer("user", identity);
-                    self.im.user.set_answer("contact", identity);
-                }
-            });
         };
 
         // override normal state adding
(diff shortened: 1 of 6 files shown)


NicolaasDieBaas:
@imsickofmaps please review

src/ussd_nurse.js
```diff
@@ -1,21 +1,113 @@
...
+            // initialising services
+            is = new IdentityStore(
+                new JsonApi(self.im, {}),
+                self.im.config.services.identity_store.token,
+                self.im.config.services.identity_store.url
+            );
+
+            var msisdn = utils.normalize_msisdn(self.im.user.addr, '27');
+            return is
+                .get_or_create_identity({"msisdn": msisdn})
```
gsvr: I think we should standardize on our indentation - this is a bit more readable but the aligned option saves horizontal space and the brackets align better, so I'd prefer using that way.

gsvr:
Please remove all of the `api.contacts.add` blocks from the testing (or comment it out for now if you want to refer to it later):
https://github.com/praekeltfoundation/ndoh-jsbox/pull/23/files#diff-00c95e8ce08e3ad8da9fdcd4dc903412R91

go-app-ussd_nurse.js
```diff
@@ -3,22 +3,114 @@ go;
...
         self.init = function() {
+            // initialising services
+            is = new IdentityStore(
+                new JsonApi(self.im, {}),
+                self.im.config.services.identity_store.token,
+                self.im.config.services.identity_store.url
+            );
+
+            var msisdn = utils.normalize_msisdn(self.im.user.addr, '27');
+            return is
```
gsvr: I don't think we should put this stuff (starting with the `return`) in `init` - I think `init` will run everytime you switch states, so it'll make multiple api calls.

test/fixtures.js
```diff
@@ -2582,7 +2589,158 @@ module.exports = function() {
...
+                        "version": 1,
+                        "details": {
+                            "default_addr_type": "msisdn",
+                            "addresses": {
+                                "msisdn": {
+                                    "+27821231111": {}
+                                }
+                            },
+                        },
+                        "nc_working_on": "+27821232222",
```
gsvr: Put this inside `details`, inside a sub-object called `nurseconnect`, and call it `working_on`, since the `nc_` would then be redundant.

test/fixtures.js
```diff
@@ -2582,7 +2589,158 @@ module.exports = function() {
...
+                            },
+                        },
+                        "nc_last_reg_id": "7",
+                        "nc_is_registered": 'true',
+                        "nc_faccode": '123456',
+                        "nc_facname": 'WCL clinic',
+                        "nc_working_on": "",
+                        "nc_id_type": "sa_id",
+                        "nc_sa_id_no": "5101025009086",
+                        "nc_dob": "1951-01-02",
```
gsvr: Put all of these `nc_` "extras" inside the `details.nurseconnect` object.

src/ussd_nurse.js
```diff
@@ -1,21 +1,113 @@
...
-            return new EndState(name, {
-                text: $("Welcome to The Department of Health's ussd_nurse.js"),
+    // DELEGATOR START STATE
+
+        self.add('state_route', function(name) {
+            // reset working_on extra
+            self.im.user.set_answer("nc_working_on", "");
+
+            var user = self.im.user.answers.user;
+            if (user.nc_is_registered === 'true') {
```
gsvr: I think it would be better and more in line with what we've done for pmtct to look if the contact has an active nurseconnect subscription, rather than looking at the extra.

test/fixtures.js
imsickofmaps: yeah, there is no `nc_working_on` field in the model. `details` is the JSON object.

src/ussd_nurse.js
NicolaasDieBaas: @gsvr ok.  The following is the function we use with the pmtct app: 
`self.has_active_pmtct_subscription = function(id) {
            return sbm
            .list_active_subscriptions(id)
            .then(function(active_subs_response) {
                var active_subs = active_subs_response.results;
                for (var i=0; i < active_subs.length; i++) {
                    // get the subscription messageset
                    return sbm
                    .get_messageset(active_subs[i].messageset)
                    .then(function(messageset) {
                        if (messageset.short_name.indexOf('pmtct') > -1) {
                            return true;
                        }
                    });
                }
                return false;
            });
        };`
Should I use the same for nurseconnect, checking the messageset shortname for 'nurseconnect'..?

gsvr: Looks right, yes.

NicolaasDieBaas:
@gsvr please review

src/ussd_nurse.js
```diff
@@ -1,21 +1,154 @@
...
+                    }
+                } else {
+                    self.im.user.set_answer("contact", identity);
+                }
+
+                self.im.user.set_answer("user", identity);
+
+                return self.states.create('state_route');
+            });
+        });
```
gsvr: This state was a bit of a misunderstanding. Please combine state_start and state_route into one state.  It should do:
1. Clear the user answers (similar to our other apps)
2. `get_or_create_identity` for the "user" (as you've done in line 77)
3. `self.im.user.set_answer("user", identity)` (as you've done in line 82)
4. Do the routing as from line 107.
We can do the "contact" saving when we get to it in the code (i.e. when they choose to register someone else).
Does that make sense?

test/fixtures.js
```diff
@@ -2614,7 +2776,143 @@ module.exports = function() {
                 }
             },
             "response": {}
-        }
+        },
+
+        // 84: get active nurseconnect subscriptions for cb245673-aa41-4302-ac47-00000031111
```
gsvr: remove word 'nurseconnect'

test/fixtures.js
```diff
@@ -2614,7 +2776,143 @@ module.exports = function() {
...
+                            'metadata': {},
+                            'created_at': "2016-08-12T06:13:29.693272Z",
+                            'updated_at': "2016-08-12T06:13:29.693272Z"
+                        }
+                    ]
+                }
+            }
+        },
+
+        // 85: get active nurseconnect subscriptions for cb245673-aa41-4302-ac47-00000034444
```
gsvr: remove word 'nurseconnect'

test/fixtures.js
```diff
@@ -2614,7 +2776,143 @@ module.exports = function() {
...
+                            'metadata': {},
+                            'created_at': "2016-08-12T06:13:29.693272Z",
+                            'updated_at': "2016-08-12T06:13:29.693272Z"
+                        }
+                    ]
+                }
+            }
+        },
+
+        // 86: get active nurseconnect subscriptions for cb245673-aa41-4302-ac47-00000037777
```
gsvr: remove word 'nurseconnect'

src/ussd_nurse.js
NicolaasDieBaas: @gsvr makes sense...

NicolaasDieBaas:
@gsvr have another look please.

src/ussd_nurse.js
```diff
@@ -69,49 +69,27 @@ go.app = function() {
             });
         };
 
-    // START STATE
+    // DELEGATOR START STATE
+
+        self.add('state_route', function(name) {
+            // reset working_on extra
+            self.im.user.set_answer("working_on", "");
```
gsvr: Please reset _all_ the user answers.

NicolaasDieBaas:
@gsvr and how about now? You'll see I removed two tests that don't make sense

gsvr:
:+1:

PR #27, opened 2016-08-16
Change fixtures

NicolaasDieBaas:
Change fixtures for NurseConnect (so far) in the following way:
- [x] create separate fixture files according to service (keep PMTCT fixtures separate for now)
- [x] apply identity roles and helpful numbering
- [x] allocate unused fixtures slots to each service
- [x] add descriptive keys to fixtures


NicolaasDieBaas:
@gsvr please review what I've done so far.  fixtures.js still exists but will be deleted. Would appreciate any input/discussion on how the roles/numbering could be done differently; currently the identity roles are described in comments in fixtures_identity_store.js; I also put some descriptive keys in the 'headers' blocks of the identity fixtures... when we have agreement on what they should look like and where they should actually be put, i can put similar keys in the fixtures of the other services (stage-based messaging & hub) too

NicolaasDieBaas:
@gsvr oh and another thing that you'll note is that the checking of fixtures in the tests looks difficult to follow currently, i.e `utils.check_fixtures_used(api, [50, 100, 103])` (100 and 103 are IdentityStore fixtures and 50 StageBasedMessaging)  We should also discuss how to adapt `check_fixtures_used`... maybe to something like `check_fixtures_used(api, [sbm_1, is_1, is_4])` (being equivalent to `utils.check_fixtures_used(api, [50, 100, 103])`)

test/fixtures_hub.js
```diff
@@ -0,0 +1,806 @@
...
+        },
+
+        // 1:
+        {
+            'request': {
+                'method': 'GET',
+                'params': {
+                    "deprecated": "fixture"
+                },
+                'headers': {
```
gsvr: @imsickofmaps js fixtures don't check the headers, so I like to omit the headers so you don't think you're checking them when you're not - do you agree with that?

test/fixtures_identity_store.js
```diff
@@ -0,0 +1,181 @@
...
+        // 0: get identity by msisdn +27820001001
+        {
+            "request": {
+                "url": 'http://is.localhost:8001/api/v1/identities/search/',
+                "method": 'GET',
+                "params": {
+                    "details__addresses__msisdn": '+27820001001'
+                },
+                "headers": {
+                    "key": "get.identity.new.27820001001"
```
gsvr: @imsickofmaps Where would you put this `key` in the fixture? I would have put it on the same level as, and just before `"request"`. Although I do see that the headers are kept when the fixtures are loaded, so maybe we should keep it in the headers?

gsvr: @imsickofmaps I would probably have used `get.is.msisdn.27820001001` here, then you could use something like `get.is.identity.the-identity-uuid` for the uuid lookups. What do you think?

test/fixtures_identity_store.js
```diff
@@ -0,0 +1,181 @@
...
+// NURSECONNECT
+    // (+27820001001) - new identity; no subscriptions
+    // (+27820001002) - existing identity with an active MomConnect subscription (no active NurseConnect subscription)
+    // (+27820001003) - existing identity with an active NurseConnect subscription
+
+
+module.exports = function() {
+    return [
+
+        // 0: get identity by msisdn +27820001001
```
gsvr: `// 100`?

test/fixtures_identity_store.js
```diff
@@ -0,0 +1,181 @@
...
+                "code": 201,
+                "data": {
+                    "url": "http://is.localhost:8001/api/v1/identities/cb245673-aa41-4302-ac47-00000001001/",
+                    "id": "cb245673-aa41-4302-ac47-00000001001",
+                    "version": 1,
+                    "details": {
+                        "default_addr_type": "msisdn",
+                        "addresses": {
+                            "msisdn": {
+                                "+2782000101": {}
```
gsvr: `...1001`, not `...101`

test/fixtures_identity_store.js
```diff
@@ -0,0 +1,181 @@
...
+                    "optout_type": "stop",
+                    "identity": "cb245673-aa41-4302-ac47-00000000001",
+                    "reason": "unknown",
+                    "address_type": "msisdn",
+                    "address": "+27820000111",
+                    "request_source": "sms_pmtct",
+                    "requestor_source_id": "0170b7bb-978e-4b8a-35d2-662af5b6daee"
+                },
+                "headers": {
+                    "key": "post.optout.27820000111"
```
gsvr: this fixture is generally inconsistent with 0001 vs 1001 vs 0111.

test/fixtures_hub.js
```diff
@@ -0,0 +1,806 @@
...
+            },
+            "response": {}
+        },
+
+        // 1:
+        {
+            'request': {
+                'method': 'GET',
+                'params': {
+                    "deprecated": "fixture"
```
gsvr: Let's remove the params for the unused fixtures.

test/fixtures_identity_store.js
```diff
@@ -0,0 +1,181 @@
...
+                    "key": "get.identity.registered.27820001002"
+                }
+            },
+            "response": {
+                "code": 200,
+                "data": {
+                    "count": 1,
+                    "next": null,
+                    "previous": null,
+                    "results": [{
```
gsvr: I know we've done it like this in the past, but fixture readability is a problem, so rather put the start of the dict on the next line (with an indent). Apply to all.

test/ussd_nurse.test.js
```diff
@@ -230,7 +235,11 @@ describe("app", function() {
                             ].join('\n')
                         })
                         .check(function(api) {
-                            utils.check_fixtures_used(api, [62, 77, 85]);
+                            // console.log(api.http.fixtures.fixtures[20].request.headers);
+                            // maybe create new function that list table of fixture keys/descriptions against number
+                            // utils.list_fixtures  (used vs unused)
+                            utils.check_fixtures_used(api, [50, 100, 103]);
+                            // check_fixtures_used(api, [sbm_2, sbm_4, is_1]);  // or adapt to something like this...?
```
gsvr: I wouldn't bother - just change the fixture comment numbers to align with these numbers.  So sbm fixtures to start at 50 and is fixtures to start at 100.

test/fixtures_stage_based_messaging.js
```diff
@@ -0,0 +1,867 @@
...
+            },
+            "response": {
+                "code": 200,
+                "data": {
+                    "count": 1,
+                    "next": null,
+                    "previous": null,
+                    "results": [
+                        {
+                            'url': 'http://sbm.localhost:8001/api/v1/subscriptions/51fcca25-2e85-4c44-subscription-1002',
```
gsvr: This may be a bit unnecessary, but it would be nice to remove all the `.localhost:800x` parts of the urls, so it's just `http://sbm/api/...`, `http://is/api/...`, etc

test/fixtures_stage_based_messaging.js
```diff
@@ -0,0 +1,867 @@
...
+                            'id': '51fcca25-2e85-4c44-subscription-1002',
+                            'version': 1,
+                            'identity': 'cb245673-aa41-4302-ac47-00000001002',
+                            'messageset': 1,
+                            'next_sequence_number': 1,
+                            'lang': "en",
```
gsvr: Hmm. This fixture looks suspicious since `lang` is a 2-char string, not 6-char.  Let's make it look as close to reality as possible.

test/fixtures_stage_based_messaging.js
```diff
@@ -0,0 +1,867 @@
...
+                    "count": 1,
+                    "next": null,
+                    "previous": null,
+                    "results": [
+                        {
+                            'url': 'http://sbm.localhost:8001/api/v1/subscriptions/51fcca25-2e85-4c44-subscription-1003',
+                            'id': '51fcca25-2e85-4c44-subscription-1003',
+                            'version': 1,
+                            'identity': 'cb245673-aa41-4302-ac47-00000001003',
+                            'messageset': 200,
```
gsvr: For the django testing I've used certain numberings for the messagesets and schedules. For consistency it would be great if they can corellate - that way it would also be more probable that we'll identify missing sets. Please see here: https://github.com/praekeltfoundation/ndoh-hub/blob/develop/ndoh_hub/utils.py#L219. Please bookmark that file so you can refer to it in the future when more sets get added.

gsvr: Now that I look at it I can see I need to expand those numbers now. Will do it after finishing this review.

test/fixtures_identity_store.js
NicolaasDieBaas: @gsvr could change fixture numbering to be in 100's.. thought that 1000's might be more 'future-proof' in some way...

test/fixtures_identity_store.js
NicolaasDieBaas: good catch

test/fixtures_identity_store.js
NicolaasDieBaas: @gsvr think this fixture actually belongs in the PMTCT fixtures file

test/fixtures_hub.js
imsickofmaps: agreed

test/fixtures_identity_store.js
imsickofmaps: put `key` at same level as `request`

imsickofmaps: yes, got with `get.is.msisdn.27820001001` please

test/fixtures_identity_store.js
gsvr: @NicolaasDieBaas No, I mean the fixture number, not the msisdn.  Should align with the numbering in the tests.

NicolaasDieBaas: @gsvr gotcha!

NicolaasDieBaas:
@gsvr please review

gsvr:
:+1: \o/

PR #32, opened 2016-08-23
NurseConnect Change flows

NicolaasDieBaas:


NicolaasDieBaas:
@imsickofmaps please review.  Two things we'll need to look at and talk about, changing of numbers and possibly having alternate states in place of `state_block_active_subs`

NicolaasDieBaas:
@imsickofmaps @gsvr please review

src/ussd_nurse.js
```diff
@@ -65,7 +66,7 @@ go.app = function() {
                 // interrupt = false;
                 // var timeout_opts = opts || {};
                 // timeout_opts.name = name;
-                // return self.states.create('st_timed_out', timeout_opts);
+                // return self.states.create('state_timed_out', timeout_opts);
```
imsickofmaps, line 69: side note: do we have a captured todo for the timeout handling?

NicolaasDieBaas: we now have --> https://github.com/praekeltfoundation/ndoh-jsbox/issues/42

imsickofmaps:
👍  some simplification can occur when we've excluded inactive identities using the seed-jsbox-utils update that's in progress

PR #30, opened 2016-08-31
NurseConnect registration submission and id update

NicolaasDieBaas:


NicolaasDieBaas:
@imsickofmaps @gsvr please review.

src/ussd_nurse.js
```diff
@@ -377,6 +376,89 @@ go.app = function() {
...
         });
 
+        self.add('state_save_nursereg', function(name) {
+        
+            var registrant_update_info = {
+                "details": {
+                    "nurseconnect": {
+                        "facname": self.im.user.answers.registrant.details.nurseconnect.facname,
+                        "faccode": self.im.user.answers.registrant.details.nurseconnect.faccode,
+                        "is_registered": "true"
```
imsickofmaps: `true` not `"true"`

src/ussd_nurse.js
```diff
@@ -377,6 +376,89 @@ go.app = function() {
...
+
+                if (self.im.user.answers.operator.details.nurseconnect === undefined) {
+                    self.im.user.answers.operator.details.nurseconnect = {};
+                }
+
+                if (self.im.user.answers.operator.details.nurseconnect.registrees === undefined) {
+                    operator_update_info.details.nurseconnect.registrees
+                        = self.im.user.answers.registrant_msisdn;
+                } else {
+                    operator_update_info.details.nurseconnect.registrees
```
imsickofmaps: why not use an array?

src/ussd_nurse.js
```diff
@@ -377,6 +376,89 @@ go.app = function() {
             });
         });
 
+        self.add('state_save_nursereg', function(name) {
+        
+            var registrant_update_info = {
+                "details": {
```
imsickofmaps: This will destroy their addresses and anything else in the details object currently. You can't patch a `details` json field. You need to read it first and update the object.

src/ussd_nurse.js
NicolaasDieBaas: guess we could... I stuck with how things were done before

src/ussd_nurse.js
NicolaasDieBaas: @imsickofmaps with identity patch, should we pass through the whole updated identity (excl. auto fields like "url", "created_at", "updated_at", "created_by", "updated_by", "user") or  pass through only the fields that have changed..?

src/ussd_nurse.js
```diff
@@ -377,6 +376,73 @@ go.app = function() {
...
+
+            // operator.id will equal registrant.id when a self registration
+            if (self.im.user.answers.operator.id !== self.im.user.answers.registrant.id) {
+                self.im.user.answers.registrant.details.nurseconnect.registered_by = self.im.user.answers.operator_msisdn;
+
+                if (self.im.user.answers.operator.details.nurseconnect === undefined) {
+                    self.im.user.answers.operator.details.nurseconnect = {};
+                }
+
+                if (self.im.user.answers.operator.details.nurseconnect.registrees === undefined) {
```
imsickofmaps: Please drop this as discussed

src/ussd_nurse.js
```diff
@@ -377,6 +376,73 @@ go.app = function() {
...
+                    "msisdn_registrant": self.im.user.answers.registrant_msisdn,  // msisdn of the registrant
+                    "msisdn_device": self.im.user.answers.operator_msisdn,  // device msisdn
+                    "faccode": self.im.user.answers.registrant.details.nurseconnect.faccode,  // facility code
+                    "language": "eng_ZA"  // currently always eng_ZA for nurseconnect
+                }
+            };
+
+            // operator.id will equal registrant.id when a self registration
+            if (self.im.user.answers.operator.id !== self.im.user.answers.registrant.id) {
+                self.im.user.answers.registrant.details.nurseconnect.registered_by = self.im.user.answers.operator_msisdn;
```
imsickofmaps: I think this should be UUID of identity not msisdn

NicolaasDieBaas:
@imsickofmaps 2nd pass review please

imsickofmaps:
👍

PR #36, opened 2016-08-31
NurseConnect Optout submission

NicolaasDieBaas:


NicolaasDieBaas:
@imsickofmaps @gsvr please review.  not so sure what `optout_type` should be set to when doing identity optout...

NicolaasDieBaas:
review with the knowledge that there are a few optout tests that were skipped that I need to go through and activate if still required... but it seems like `state_optout` might have to change and the tests removed.  Can an user have an active subscription but be opted out..?

imsickofmaps:
👍

go-app-ussd_nurse.js
```diff
@@ -738,8 +737,35 @@ go.app = function() {
...
+                            "action": "nurse_optout",
+                            "data": {
+                                "reason": choice.value
+                            }
+                        };
+
+                        return Q
+                        .all([
+                            // optout identity
+                            is.optout(optout_info),
```
imsickofmaps: Drop this, it's unsubscribe.

imsickofmaps:
@gsvr remove optout then 👍  for when Travis checked

PR #41, opened 2016-08-31
MomConnect Clinic port - part 1

gsvr:
Flow porting. TODO's will be handled in subsequent parts.

### DIFF
--- a/src/ussd_clinic.js
+++ b/src/ussd_clinic.js
@@ -35,7 +35,7 @@ go.app = function() {
         self.number_opted_out = function(identity, msisdn) {
             var details_msisdn = identity.details.addresses.msisdn[msisdn];
             if ("optedout" in details_msisdn) {
-                return details_msisdn.optedout === true;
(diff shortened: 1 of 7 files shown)


gsvr:
@imsickofmaps Please review.

src/ussd_clinic.js
```diff
@@ -1,25 +1,479 @@
...
+        };
+
+        self.readable_sa_msisdn = function(msisdn) {
+            readable_no = '0' + msisdn.slice(msisdn.length-9, msisdn.length);
+            return readable_no;
+        },
+
+        self.number_opted_out = function(identity, msisdn) {
+            var details_msisdn = identity.details.addresses.msisdn[msisdn];
+            if ("optedout" in details_msisdn) {
```
imsickofmaps, line 37: probably safer to check if it is there that it's not `"false" || false`

imsickofmaps:
One tweak then :+1: when travis is happy

PR #47, opened 2016-09-01
MOMZA-213 Allow multiple active subs on NDOH Control

imsickofmaps:
If the old single-sub-only rule isn't enforced (it's not on Training system) the PMTCT falls over


imsickofmaps:
@NicolaasDieBaas review please

NicolaasDieBaas:
👍

PR #48, opened 2016-09-01
MomConnect Clinic port - part 2

gsvr:
- [x] post registration
- [x] update identity
- [x] send welcome sms
- [x] handle timeouts

### DIFF
--- a/src/ussd_clinic.js
+++ b/src/ussd_clinic.js
@@ -18,6 +18,7 @@ go.app = function() {
         // variables for services
         var is;
         var hub;
+        var ms;
 
         self.init = function() {
             // initialise services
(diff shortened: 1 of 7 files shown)


src/ussd_clinic.js
```diff
@@ -84,14 +97,112 @@ go.app = function() {
             }
         };
 
+        self.compile_registrant_info = function() {
+            var registrant_info = self.im.user.answers.registrant;
+            registrant_info.details.lang_code = self.im.user.answers.lang_code;
+            registrant_info.details.consent =
```
imsickofmaps, line 103: can we make this `true` not `"true"`?

imsickofmaps:
Minor tweak then :+1: when travis is happy

src/ussd_clinic.js
gsvr: Yes, but I'm opening issue #51 for it since it's accross different apps that have pull requests open at the moment.

PR #52, opened 2016-09-05
Optimise active subscription finding

gsvr:
Stop returning promises in a for loop, rather get all the messagesets as well and compare the id's.


gsvr:
@imsickofmaps Please review. Travis will fail since `seed-jsbox-utils` doesn't have the list_messagesets function yet. If you're happy with the principle, we should perhaps move the function `has_active_subscription` to `seed-jsbox-utils`.

imsickofmaps:
👍

PR #72, opened 2016-09-22
PMTCT - fix get_january call

NicolaasDieBaas:
a `make_month_choices` call inside `state_birth_month` is being made.
As one of `make_month_choices` arguments, do a `utils.get_january` call instead of the current `get_january`

Things seem to work anyway, but think it's better to be more clear in calling `get_january`


NicolaasDieBaas:
@gsvr please review

gsvr (approved):

PR #83, opened 2017-01-13
Fix sending edd for pmtct registration

rudigiesler:
At the moment, the app is looking at `identity.last_edd` for the edd date, when it is actually being stored at `identity.details.last_edd`. We should change it to look there, and change the tests to actually check that it is sending the correct information to the registration service.


rudigiesler:
Ready for review.

erikh360 (approved):

PR #89, opened 2017-05-05
Feature/MOMZA-222-USSD-POPI-FAQ-Service

CarlaLlama:


src/ussd_faq.js
```diff
@@ -0,0 +1,329 @@
...
+                // Average sessions to register
+                .add.tracker({
+                    action: 'exit',
+                    state: 'state_start'
+                }, {
+                    action: 'enter',
+                    state: 'state_end_success'
+                }, {
+                    sessions_between_states: [self.metric_prefix, 'avg.sessions_to_register'].join('.')
+                })
```
KaitCrawford: You probably don't need this metric. It's intended to track how many sessions it takes a mom to register for momconnect which isn't really relevant here.

src/ussd_popi_faq.js
```diff
@@ -0,0 +1,329 @@
...
+                // Average sessions to register
+                .add.tracker({
+                    action: 'exit',
+                    state: 'state_start'
+                }, {
+                    action: 'enter',
+                    state: 'state_end_success'
+                }, {
+                    sessions_between_states: [self.metric_prefix, 'avg.sessions_to_register'].join('.')
+                })
```
KaitCrawford: You probably don't need this metric. It's intended to track the number of sessions it takes for a user to register for momconnect so isn't really relevant for this.

src/ussd_popi_faq.js
```diff
@@ -0,0 +1,329 @@
...
+                    new Choice(creator_opts.name, $('Continue viewing FAQ')),
+                    new Choice('state_start', $('Main menu'))
+                ],
+                next: function(choice) {
+                    return choice.value;
+                }
+            });
+        });
+
+        self.add("state_start", function(name) {  // interstitial state
```
KaitCrawford: You've pulled a lot from the ussd registration jsbox into here and it's over-complicated this state a bit. maybe use the sms jsbox for reference instead since that's also intended for users who have already registered https://github.com/praekeltfoundation/ndoh-jsbox/blob/develop/src/sms_inbound.js#L115

src/ussd_popi_faq.js
```diff
@@ -0,0 +1,329 @@
...
+                new JsonApi(self.im, {}),
+                self.im.config.services.identity_store.token,
+                self.im.config.services.identity_store.url
+            );
+            sbm = new SeedJsboxUtils.StageBasedMessaging(
+                new JsonApi(self.im, {}),
+                self.im.config.services.stage_based_messaging.token,
+                self.im.config.services.stage_based_messaging.url
+            );
+            hub = new SeedJsboxUtils.Hub(
```
KaitCrawford: It doesn't look like you use the Hub at all (this is where we store registrations) so you don't need to set up this connection to it

src/ussd_popi_faq.js
```diff
@@ -0,0 +1,329 @@
...
+                new JsonApi(self.im, {}),
+                self.im.config.services.hub.token,
+                self.im.config.services.hub.url
+            );
+            ms = new SeedJsboxUtils.MessageSender(
```
KaitCrawford: The same goes for this connection to the MessageSender (this is what sends out sms messages)

src/ussd_popi_faq.js
```diff
@@ -0,0 +1,329 @@
...
+                    state: 'state_start'
+                }, {
+                    action: 'enter',
+                    state: 'state_end_success'
+                }, {
+                    sessions_between_states: [self.metric_prefix, 'avg.sessions_to_register'].join('.')
+                })
+            ;
+
+            self.im.on('state:exit', function(e) {
```
KaitCrawford: you probably don't need this method or `fire_complete`. It just fires a metric to record how many people completed the registration. Since we never exit and it's up to the user to close the session, I don't think this should ever get called

KaitCrawford (commented):
Looking great. I've left a few comments around stuff that doesn't need to be there

test/ussd_popi_faq.test.js
```diff
@@ -0,0 +1,597 @@
+var vumigo = require("vumigo_v02");
+var AppTester = vumigo.AppTester;
+var assert = require('assert');
+var fixtures_IdentityStore = require('./fixtures_identity_store');
+var fixtures_StageBasedMessaging = require('./fixtures_stage_based_messaging');
+var fixtures_MessageSender = require('./fixtures_message_sender');
+var fixtures_Hub = require('./fixtures_hub');
+var fixtures_Jembi = require('./fixtures_jembi');
+var fixtures_ServiceRating = require('./fixtures_service_rating');
```
KaitCrawford: Take a look at these fixtures and make sure you need all of them. I suspect there are a few that can be removed

test/ussd_popi_faq.test.js
```diff
@@ -0,0 +1,597 @@
...
+                    name: 'ussd_popi_faq',
+                    env: 'test',
+                    metric_store: 'test_metric_store',
+                    testing_today: "2014-04-04 07:07:07",
+                    testing_message_id: '0170b7bb-978e-4b8a-35d2-662af5b6daee',
+                    logging: "off",
+                    no_timeout_redirects: [
+                        "state_start", "state_end_not_pregnant", "state_end_consent_refused",
+                        "state_end_success", "state_registered_full", "state_registered_not_full",
+                        "state_end_compliment", "state_end_complaint", "state_end_go_clinic"],
```
KaitCrawford: You can remove a bunch of states from here since they don't exist in this app. This is a list of states that shouldn't redirect to the `state_timed_out` if a gets disconnected and redials. I think you'll only need `start_state` here.

test/ussd_popi_faq.test.js
```diff
@@ -0,0 +1,597 @@
...
+                            url: 'http://sbm/api/v1/',
+                            token: 'test StageBasedMessaging'
+                        },
+                        hub: {
+                            url: 'http://hub/api/v1/',
+                            token: 'test Hub'
+                        },
+                        message_sender: {
+                            url: 'http://ms/api/v1/',
+                            token: 'test MessageSender'
```
KaitCrawford: You shouldn't need the hub and message_sender connection details here either.

KaitCrawford (commented):
Looking great! A few things to remove from the tests but good job!

test/ussd_popi_faq.test.js
CarlaLlama: That makes sense. Thanks! I wasn't sure what no_timeout_redirect meant :)

test/ussd_popi_faq.test.js
CarlaLlama: Thanks! I should have removed this when I removed it from the js file *facepalm*

KaitCrawford (approved):
:+1:  Awesome work!

PR #96, opened 2017-05-10
Additional USSD service for POPI

CarlaLlama:


src/ussd_popi_user_data.js
```diff
@@ -0,0 +1,333 @@
...
+                        self.im.user.set_answer("msisdn", null);
+                        return 'state_info_deleted';
+                    } else {
+                        return 'state_info_not_deleted';
+                    }
+                }
+            });
+        });
+
+                self.add('state_select_language', function(name) {
```
KaitCrawford: The indentation here is a little off :smile:

test/ussd_popi_user_data.test.js
```diff
@@ -0,0 +1,354 @@
+var vumigo = require("vumigo_v02");
+var AppTester = vumigo.AppTester;
+var assert = require('assert');
+var fixtures_IdentityStore = require('./fixtures_identity_store');
+var fixtures_StageBasedMessaging = require('./fixtures_stage_based_messaging');
+var fixtures_ServiceRating = require('./fixtures_service_rating');
```
KaitCrawford: I don't think this is required since you don't use any of the fixtures from it.

test/ussd_popi_user_data.test.js
```diff
@@ -0,0 +1,354 @@
...
+                           
+                            .check.interaction({
+                                state: "state_info_deleted",
+                                reply: [
+                                    'Thank you. Your personal information stored on ' +
+                                    'MomConnect has been permanently removed.'
+                                ].join('\n')
+                            })
+                            .check(function(api) {
+                                // later check that actually has been removed
```
KaitCrawford: A common format for these sorts of comments is `// TODO: check that actually has been removed`
Some IDEs will find those comments and make a To Do list from them

src/ussd_popi_user_data.js
```diff
@@ -0,0 +1,432 @@
...
+                return sbm
+                // check that user is registered on momconnect
+                .check_identity_subscribed(self.im.user.answers.operator.id, "momconnect")
+                .then(function(identity_subscribed_to_momconnect) {
+                    if (identity_subscribed_to_momconnect) {
+                        var mset = sbm.list_active_subscriptions();
+                        var allmset = '';
+                        for(var i = 0; i < mset.length; i++){
+                            allmset += sbm.get_messageset(mset[i]);
+                        }
```
KaitCrawford: I don't think this will work. `list_active_subscriptions` returns something that looks like
`{ "count": 1,
    "next": null,
    "previous": null,
    "results": [
          {...}
     ]
}`
So the actual list of subscriptions is in the `results` attribute of the object that gets returned.
when you loop through those subscriptions you then need to get the `messageset` attribute from each and send that to `get_messageset` (right now you'd be sending the whole subscription object). Then for each messageset you'd want to get the `short_name` attribute and add that to the message sets. It looks like there isn't an existing fixture for that call to get the messageset so I'll give you access to one of our QA systems so you can see what that would look like.

test/ussd_popi_user_data.test.js
```diff
@@ -0,0 +1,630 @@
...
+                            })
+                            .check(function(api) {
+                                utils.check_fixtures_used(api, [51, 54, 141, 171]);
+                            })
+                            .check.reply.ends_session()
+                            .run();
+                        }); 
+                    });
+                });
+
```
KaitCrawford, line 181: You should probably add two tests similar to the above ones but with the messageset info in the output.

src/ussd_popi_user_data.js
CarlaLlama: Ah, that makes sense. Thanks!

CarlaLlama:
@KaitCrawford finally fixed it!

go-app-ussd_popi_user_data.js
```diff
@@ -185,19 +185,27 @@ go.app = function() {
...
-                        var allmset = '';
-                        for(var i = 0; i < mset.length; i++){
-                            allmset += sbm.get_messageset(mset[i]);
-                        }
-                        self.im.user.set_answer("message_sets",allmset);
-                        return self.states.create('state_all_options_view');
+                        
+                        return sbm
+                        .list_active_subscriptions(self.im.user.answers.operator.id)
+                        .then(function(active_subscriptions){
```
KaitCrawford: I think you'll need to loop through the active subscriptions here. Right now you'll only get the message set for one of their subscriptions. It's unlikely but possible that a user might have multiple subscriptions.

KaitCrawford (commented):
looking great :smile:

CarlaLlama:
@KaitCrawford Okay! I think it's done this time xD

test/ussd_popi_user_data.test.js
KaitCrawford: Do you need these fixtures?

src/ussd_popi_user_data.js
```diff
@@ -143,6 +145,10 @@ go.app = function() {
                 }
         };
 
+        self.return_messageset = function(active_subscriptions, i){
```
KaitCrawford: I think this might be left over from something you tried.

src/ussd_popi_user_data.js
```diff
@@ -216,26 +225,12 @@ go.app = function() {
                 question: $('What would you like to do?'),
                 choices: [
                     new Choice('state_view', $('See my personal info')),
+                    new Choice('state_view_sms', $('Send my personal info as sms')),
```
KaitCrawford: Did you intend to put this state back into the flow?

src/ussd_popi_user_data.js
```diff
@@ -1,3 +1,4 @@
+/*jshint loopfunc:true */
```
KaitCrawford: What is this comment for?

package.json
```diff
@@ -26,7 +26,8 @@
     "lodash": "~2.4.1",
     "q": "~0.9.7",
     "moment": "~2.6",
-    "go-jsbox-metrics-helper": "^0.1.0"
+    "go-jsbox-metrics-helper": "^0.1.0",
+	"es6-promise": "~3.1.2"
```
KaitCrawford: Just correct this indentation please :)

KaitCrawford: Please fix this indentation :D

KaitCrawford (commented):
@CarlaLlama this is looking great :smile:
I just have a few comments, mostly one or two lines that I'm not sure if they should be there.

src/ussd_popi_user_data.js
CarlaLlama: That suppresses the warning when a promise is made in a loop. But I think I can remove it now actually..

src/ussd_popi_user_data.js
CarlaLlama: Yeah, whoops.. Thanks!

src/ussd_popi_user_data.js
CarlaLlama: Yes. I ran out of options because I needed a paginated state for state_view, but paginated state only gives you the options for More and Exit, so I couldn't add the option to Send As Sms

package.json
CarlaLlama: Sure!

src/ussd_popi_user_data.js
```diff
@@ -0,0 +1,445 @@
...
+                    if (identity_subscribed_to_momconnect) {
+                        var promises = [];
+                        return sbm
+                        .list_active_subscriptions(self.im.user.answers.operator.id)
+                        .then(function(active_subscriptions){
+                            promises = active_subscriptions.results.map(function(result){
+                                return sbm.get_messageset(result.messageset); 
+                            });
+                            var sets = '';
+                            return Promise.all(promises)
```
rudigiesler: Because the javascript sandbox was written in an older version of node, and still runs using that older version, `Promise` didn't exist, so we used the `Q` library, which has the same functions: http://documentup.com/kriskowal/q/#combination . I think we can use that instead of having to include the extra dependancy.

Sorry for the confusion around this.

CarlaLlama: Oh! I was wondering what Q was. Thanks, I'll fix that :)

KaitCrawford (approved):
:+1: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake: :cake:

PR #97, opened 2017-05-22
Updated operating times from 8am-4pm to 8am-6pm

CarlaLlama:


KaitCrawford (approved):
:+1:
If this is something that's required sooner rather than later we should probably look into the translation process

PR #100, opened 2017-06-01
Implement language change in user data jsbox

KaitCrawford:


src/ussd_popi_user_data.js
```diff
@@ -316,16 +322,35 @@ go.app = function() {
...
             });
         });
 
+        self.add('state_switch_lang', function(name) {
+            var change_info = {
+                "registrant_id": self.im.user.answers.operator.id,
+                "action": "momconnect_change_language",
+                "data": {
+                    "language": self.im.user.answers.new_language,
+                    "old_language": self.im.user.answers.operator.details.lang_code
```
rudigiesler: Should we make this a required field on the Hub?

go-app-ussd_popi_user_data.js
```diff
@@ -319,16 +325,35 @@ go.app = function() {
...
+            var change_info = {
+                "registrant_id": self.im.user.answers.operator.id,
+                "action": "momconnect_change_language",
+                "data": {
+                    "language": self.im.user.answers.new_language,
+                    "old_language": self.im.user.answers.operator.details.lang_code
+                }
+            };
+
+            self.im.user.set_answer("state_language", self.im.user.answers.new_language);
```
rudigiesler: It doesn't seem like we use this answer for anything?

go-app-ussd_popi_user_data.js
```diff
@@ -319,16 +325,35 @@ go.app = function() {
...
                 },
             });
         });
 
+        self.add('state_switch_lang', function(name) {
+            var change_info = {
+                "registrant_id": self.im.user.answers.operator.id,
+                "action": "momconnect_change_language",
+                "data": {
+                    "language": self.im.user.answers.new_language,
```
rudigiesler: We can probably just use the `state_select_language` answer here, and then we don't need to set `new_language` in that state either.

src/ussd_popi_user_data.js
KaitCrawford: yeah, I think so. That way we always know what language they changed from.

go-app-ussd_popi_user_data.js
KaitCrawford: ah ok, I though you had to explicitly set the answers for them to persist through states.

src/ussd_popi_user_data.js
rudigiesler: Since we know the old language in the hub before we change it, I think I'll just make it add the old language in the hub (so add that field when we do the change, that way it doesn't matter whether or not the app supplies that field, so it doesn't need to be required)

rudigiesler: I've created a pull request for this here: https://github.com/praekeltfoundation/ndoh-hub/pull/101

go-app-ussd_popi_user_data.js
```diff
@@ -319,16 +325,33 @@ go.app = function() {
...
+            var change_info = {
+                "registrant_id": self.im.user.answers.operator.id,
+                "action": "momconnect_change_language",
+                "data": {
+                    "language": self.im.user.answers.state_select_language,
+                    "old_language": self.im.user.answers.operator.details.lang_code
+                }
+            };
+
+            self.im.user.set_lang(self.im.user.answers.state_select_language);
```
rudigiesler: Sorry that I didn't see this before, but it seems like we're calling `set_lang` already in the previous state? Maybe we can remove all the `next` logic from the previous state, and just go directly to this state, which does all the logic to change the correct things?

Also, `set_lang` returns a promise, so by not waiting for that promise to be fulfilled before moving on, we don't have any guarantees that the user's language will be change by the next state, so sometimes they will get the next state in the new language, and sometimes in the old language.

rudigiesler (approved):

PR #101, opened 2017-06-02
Implement number change in user data jsbox

KaitCrawford:
The diff is a bit weird because I changed the indentation on two of the existing tests :/

### DIFF
--- a/src/ussd_popi_user_data.js
+++ b/src/ussd_popi_user_data.js
@@ -275,7 +275,7 @@ go.app = function() {
                 choices: [
                     new Choice('state_select_language', $('Update my language choice')),
                     new Choice('state_change_identity', $('Update my identification')),
-                    new Choice('state_change_msisdn', $('Use a different phone number'))
+                    new Choice('state_new_msisdn', $('Use a different phone number'))
                 ],
                 next: function(choice) {
                     return choice.value;
@@ -408,7 +408,7 @@ go.app = function() {
             });
         });
 
-        self.add('state_change_msisdn', function(name) {
+        self.add('state_new_msisdn', function(name) {
             return new FreeText(name, {
                 question: $('Please enter the new phone number we should use ' +
                             'to send you messages eg. 0813547654'),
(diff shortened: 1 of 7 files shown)


src/ussd_popi_user_data.js
```diff
@@ -420,32 +420,112 @@ go.app = function() {
...
+            .then(function(identities_found) {
+                var identities = (identities_found.results.length > 0) ? identities_found.results : null;
+
+                if (identities === null) {
+                    return self.states.create('state_create_msisdn_change');
+                }
+
+                var opted_out_on_operator = false;
+                var msisdn_available = true;
+                for(i=0; i < identities.length; i++) {
```
rudigiesler: We can probably use lodash's forEach here, instead of the `for(i=...) {identities[i]...}` pattern here, to make things a little cleaner: https://lodash.com/docs/4.17.4#forEach

src/ussd_popi_user_data.js
```diff
@@ -420,32 +420,112 @@ go.app = function() {
...
+                      if (identities[i].details.addresses.msisdn[new_msisdn].optedout ||
+                              identities[i].details.addresses.msisdn[new_msisdn].inactive)
+                        opted_out_on_operator = true;
+                    }
+                    // Msisdn active on any other than this operator
+                    if (!identities[i].details.addresses.msisdn[new_msisdn].optedout &&
+                            !identities[i].details.addresses.msisdn[new_msisdn].inactive) {
+                        if (identities[i].id !== self.im.user.answers.operator.id) {
+                            msisdn_available = false;
+                            break;
```
rudigiesler: I think here we can just do a `return self.states.create('state_msisdn_change_fail');`, and not have to deal with the `msisdn_available` variable.

src/ussd_popi_user_data.js
```diff
@@ -420,32 +420,112 @@ go.app = function() {
...
+                }
+
+                var opted_out_on_operator = false;
+                var msisdn_available = true;
+                for(i=0; i < identities.length; i++) {
+                    // Msisdn already on this operator but inactive
+                    if (identities[i].id === self.im.user.answers.operator.id) {
+                      if (identities[i].details.addresses.msisdn[new_msisdn].optedout ||
+                              identities[i].details.addresses.msisdn[new_msisdn].inactive)
+                        opted_out_on_operator = true;
```
rudigiesler: It's a small thing, but I prefer not to use the one line if statement things, and to be explicit with using `{}` on all of the if statements. It makes it more explicit, and easier to add on at a later stage.

src/ussd_popi_user_data.js
```diff
@@ -420,32 +420,112 @@ go.app = function() {
...
+
+        self.states.add("state_opt_in", function(name) {
+            var optin_info = {
+                "identity": self.im.user.answers.operator.id,
+                "address_type": "msisdn",
+                "address": self.im.user.answers.new_msisdn,
+                "request_source": self.im.config.name || "ussd_popi_user_data"
+            };
+            return is
+            .optin(optin_info)
```
rudigiesler, line 503: Would this optin also set the default value if required?

src/ussd_popi_user_data.js
KaitCrawford: This is actually a product of forgetting to put any parenthesis around the `if` blocks and only fixing the ones that caused the tests to fail :sweat_smile:

src/ussd_popi_user_data.js
KaitCrawford: Hmmm good point. Investigation says not (https://github.com/praekelt/seed-identity-store/blob/develop/identities/models.py#L126).
I'll get it to create a change as well so that it gets set as the default.

src/ussd_popi_user_data.js
KaitCrawford: I'm struggling to get this to work when the code is wrapped in a function so it can be used by the forEach()

rudigiesler: Ah right, that's because a `forEach` returns a list of all the return values, so it won't exit immediately.

rudigiesler (approved):

PR #102, opened 2017-06-02
Implement ability to change identification in user data app

KaitCrawford:


go-app-ussd_popi_user_data.js
```diff
@@ -375,39 +375,67 @@ go.app = function() {
...
-                    new Choice('state_change_passport_drc', $('DRC')),
-                    new Choice('state_change_passport_som', $('Somalia')),
-                    new Choice('state_change_passport_oth', $('Other'))
+                    new Choice('zw', $('Zimbabwe')),
+                    new Choice('mz', $('Mozambique')),
+                    new Choice('mw', $('Malawi')),
+                    new Choice('ng', $('Nigeria')),
+                    new Choice('cd', $('DRC')),
+                    new Choice('so', $('Somalia')),
+                    new Choice('other', $('Other'))
```
rudigiesler, line 387: In the current ndoh_hub setup, we have a different list of countries: https://github.com/praekeltfoundation/ndoh-hub/blob/develop/ndoh_hub/utils.py#L79 . I'm not sure whether that list of countries needs to change, or whether we need to change the list of countries in the hub app.

src/ussd_popi_user_data.js
```diff
@@ -372,39 +372,67 @@ go.app = function() {
...
+                    new Choice('mw', $('Malawi')),
+                    new Choice('ng', $('Nigeria')),
+                    new Choice('cd', $('DRC')),
+                    new Choice('so', $('Somalia')),
+                    new Choice('other', $('Other'))
                 ],
                 next: function(choice) {
-                    // TODO: Update passport on identitystore
-                    return 'state_update_passport';
+                    return 'state_passport_no';
```
rudigiesler: Not strictly necessary, but just for future, `next` can just be a string, so you don't have to have a function that returns the string, you can just have the value as the string: http://vumi-jssandbox-toolkit.readthedocs.io/en/master/states/reference.html#ChoiceState

rudigiesler (commented):
Just one comment from me, not sure if it needs action here or somewhere else.

go-app-ussd_popi_user_data.js
KaitCrawford: @rudigiesler well spotted! It looks like the nurseconnect and momconnect lines use different lists and the list in the hub app is only for the nurseconnect line. It looks to me like momconnect registrations for any countries other than 'mz' should have been marked as invalid :disappointed:
I'll extend the list of valid countries to include those in the momconnect list. We may want to investigate the work required to fix registrations that have been affected by this and fast-forward the mothers to the correct stage.

rudigiesler: I'll run some checks to get some numbers on the amount of affected registrations, and come up with a plan

rudigiesler (approved):

PR #103, opened 2017-06-05
Implement optout option in user data ussd line

KaitCrawford:


test/ussd_popi_user_data.test.js
```diff
@@ -689,10 +689,7 @@ describe('app', function() {
                                 ].join('\n')
                             })
                             .check(function(api) {
-                                // TODO: check that actually has been removed
```
rudigiesler: Maybe here we should check that we have actually deleted the `operator` and `msisdn` fields off of the contact's answers?

rudigiesler (commented):
Just one comment from me :)

rudigiesler (commented):
Looks good to me.

I'm just wondering if we might need to tell the hub as well? So that it can be posted to Jembi, and so that we can cancel all subscriptions?

KaitCrawford:
Oh right, I keep forgetting that the hub doesn't subscribe to the optout hook on the identity store. I'll make this go through the hub then rather.

rudigiesler (approved):

PR #108, opened 2017-06-14
Set language according to identity store

rudigiesler:


erikh360 (commented):
Shouldn't you run `npm test` to generate the file in the root dir?

Is it possible to write a test to verify this?

rudigiesler:
@erikh360 Yes you're correct, I've updated with the generated file. I've also added a test for this. Ready for another review.

erikh360 (approved):
🍰

PR #109, opened 2017-06-16
Allow channel to be set per message.

smn:
Depending on a user's preferences, which may no longer be around in
self.users.answers state, we should be able to select a different 
channel to send messages.

The welcome message state would still be around.
The compliments & complaints state will likely have expired, this
requires a lookup.

This PR requires the changes in https://github.com/praekelt/seed-jsbox-utils/pull/40

### DIFF
--- a/test/fixtures_pilot.js
+++ b/test/fixtures_pilot.js
@@ -62,8 +62,6 @@ module.exports = function() {
             }
         },
         post_outbound_message: function(params) {
-            // defaulting to this identity + address as its used in many
-            // fixtures
             params = params || {};
             var identity = params.identity;
             var address = params.address;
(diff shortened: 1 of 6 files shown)


smn:
@miltontony @erikh360 @rudigiesler ready for review but obviously requires praekelt/seed-jsbox-utils#40 to get merged in and released before Travis tests will pass.

test/fixtures_pilot.js
```diff
@@ -64,6 +60,81 @@ module.exports = function() {
                     "data": {}
                 }
             }
-        }
+        },
+        post_outbound_message: function(params) {
+            // defaulting to this identity + address as its used in many
```
rudigiesler: This comment says that we're defaulting to a specific identity and address, but below it looks like we're not defaulting to a specific identity and address.

smn:
@rudigiesler thanks! removed that docstring

rudigiesler (approved):

PR #110, opened 2017-06-19
Use local state if available to determine channel

smn:


smn:
@miltontony ready for review

miltontony:
👍

PR #112, opened 2017-06-28
updated translations, now with WhatsApp

smn:


smn:
@rudigiesler & @KaitCrawford translations updated to include WhatsApp. I've extracted the strings from the ussd app into a new template (.pot) and then updated the .pos from that template. It looks like that hasn't been done in a long time (2014 or 2016 was the latest update) and so quite a few translation strings have been removed. The previous template seems to have been for a file called `go-app-personal.js` which seems ancient.

smn:
As a result this is going to be hard to review, I've essentially only added a new translation string for WhatsApp: 
```
How would you like to receive messages about you and your baby? 1. WhatsApp 2. SMS
```

KaitCrawford (approved):
:+1:

PR #119, opened 2017-08-04
Replace check_identity_subscribed call

alexmuller:
This is deprecated and should be replaced with `is_identity_subscribed`.

In the version of seed-jsbox-utils that we're using, they have almost
identical behaviour:

https://github.com/praekelt/seed-jsbox-utils/blob/0.1.17/lib/stage_based_messaging.js#L162-L189


KaitCrawford (approved):
:+1:

PR #121, opened 2017-08-08
Remove sms option and timeout redirects

KaitCrawford:


go-app-ussd_popi_faq.js
```diff
@@ -63,32 +61,7 @@ go.app = function() {
...
-                question: $('Welcome back. Please select an option:'),
-                choices: [
-                    new Choice(creator_opts.name, $('Continue viewing FAQ')),
-                    new Choice('state_start', $('Main menu'))
-                ],
-                next: function(choice) {
-                    return choice.value;
-                }
-            });
-        });
```
rudigiesler: So the reason we have these timeout states is although we can have 182 (I think?) chars per ussd message, some providers have additional information in the first message, so the first message of a session has a lower limit (I think around 140 chars).

I just want to make sure that we've checked that none of the screens or their translations are over the 140 char limit.

The other option (which I think might be better for the FAQ, since it's not a long registration flow), is to redirect them to the main menu every time they redial after timeout.

go-app-ussd_popi_user_data.js
```diff
@@ -155,33 +138,7 @@ go.app = function() {
...
-            return new ChoiceState(name, {
-                question: $('Welcome back. Please select an option:'),
-                choices: [
-                    new Choice('state_start', $('Main menu'))
-                ],
-                next: function(choice) {
-                    return choice.value;
-                }
-            });
-        });
```
rudigiesler: Same question here with the 182/140 character limit and whether we've checked text + translations.

go-app-ussd_popi_user_data.js
```diff
@@ -226,32 +183,23 @@ go.app = function() {
...
-                    if (choice.value === 'state_view_sms') {
-                        return self
-                        .send_data_as_sms()
-                        .then(function() {
-                            return 'state_view_sms';
-                        });
-                    }else{
-                        return choice.value;
-                    }
+                    return choice.value;
```
rudigiesler: We can probably now change this to a MenuState and remove this logic: http://vumi-jssandbox-toolkit.readthedocs.io/en/master/states/reference.html#MenuState

go-app-ussd_popi_faq.js
KaitCrawford: @rudigiesler for your second point there: wouldn't removing the timeout redirects cause them to hit the main menu every time they dial anyway?

KaitCrawford: @rudigiesler I'm confused as to why the reduced space in the first message would require us to have the timeout states? because they'd have more screens to traverse?

rudigiesler: Removing the timeout states means that they'll go back to whatever state they were when they timed out.

So if they were busy viewing a long screen of information, and then they timeout, and then redial in again, they'll be shown that long screen as the first thing.

The idea with the timeout state is that when they redial, they'll be shown that timeout state instead, which we make sure is <140 chars. They then click continue, and see whatever screen they're on. That way we ensure that the first screen of every session is <140 chars.

rudigiesler (approved):

rudigiesler (approved):

PR #122, opened 2017-08-11
Prevent duplicate clinic registrations

alexmuller:
This is a first iteration of what a solution to duplicate clinic registrations might look like.

There may be some edge cases that I haven't considered.

If this looks ok I'll add some tests for it before merging.

### DIFF
--- a/go-app-ussd_clinic.js
+++ b/go-app-ussd_clinic.js
@@ -158,6 +158,7 @@ go.app = function() {
     var FreeText = vumigo.states.FreeText;
     var EndState = vumigo.states.EndState;
     var ChoiceState = vumigo.states.ChoiceState;
+    var MenuState = vumigo.states.MenuState;
     var PaginatedChoiceState = vumigo.states.PaginatedChoiceState;
     var Choice = vumigo.states.Choice;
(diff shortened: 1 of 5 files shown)


rudigiesler:
Maybe we should be offering a choice to type in a different number, rather than just ending the registration?

Otherwise looks good to me

alexmuller:
@rudigiesler Thanks, have done! I've got most of the tests working but there are now just a couple I can't figure out. I'm not sure why fixture 54 (messagesets) is now being used.

alexmuller:
All sorted now with help from Rudi! Before deployment this will need a token added to the config repo.

alexmuller:
Oh also what's the situation with localisation in this repo? Do we need translations for this text?

rudigiesler:
@alexmuller Yes we do need translations. You will find them in the `config` folder. They usually generate large diffs, so we usually do them in a separate PR.

alexmuller:
Updated to fix @KaitCrawford's feedback that we also need to do this check at `state_mobile_no`. Finding this repo really hard to make changes to 😞 

Also need to do translations of the new text.

KaitCrawford (approved):
:+1:

PR #125, opened 2017-08-29
[MOMZA-766] Prevent duplicate registrations for ussd_chw

alexmuller:
This is very similar to #122 except for chw (community health workers?) rather than clinics. The fixtures were a lot more challenging this time.


alexmuller:
Ah I think this will also need a new token in the config repo before merging.

KaitCrawford (approved):
:+1:

PR #128, opened 2017-09-04
Fix prebirth subscription test

alexmuller:
@KaitCrawford pointed out that we also have a `prebirth.hw_partial` message set and we do want somebody to be able to have a partial and a full subscription at the same time.

The string to match is a regex so we only want to match a literal dot: `\.`

### DIFF
--- a/go-app-ussd_clinic.js
+++ b/go-app-ussd_clinic.js
@@ -508,7 +508,7 @@ go.app = function() {
                 self.im.user.set_answer("operator", identity);
                 self.im.user.set_answer("operator_msisdn", operator_msisdn);
 
-                return sbm.is_identity_subscribed(identity.id, [/prebirth/]);
+                return sbm.is_identity_subscribed(identity.id, [/prebirth\.hw_full/]);
             })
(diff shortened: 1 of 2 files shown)


src/ussd_chw.js
```diff
@@ -286,7 +286,7 @@ go.app = function() {
                 self.im.user.set_answer("operator", identity);
                 self.im.user.set_answer("operator_msisdn", operator_msisdn);
 
-                return sbm.is_identity_subscribed(identity.id, [/prebirth/]);
+                return sbm.is_identity_subscribed(identity.id, [/prebirth\.hw_full/]);
```
KaitCrawford: You can leave this one as it was. The chw line (for 'community health worker') just subscribes them to a set with the same content as the public line.

KaitCrawford (commented):
I don't think you need to change the chw app but I do think you probably need to change the `state_mobile_no` in the clinic app as well

alexmuller:
@KaitCrawford Ah right thanks, updated!

KaitCrawford (approved):
:+1:

PR #130, opened 2017-09-14
Optimize the WhatsApp number checks

smn:
1. Specify the number we're wanting to check for.

If we don't specify the number, all numbers for the account are checked
and that generates more load than necessary.

2. In the public registration line, make 2 separate calls (like we do for clinics)

The first call has "wait=False" which forks the WhatsApp query to the background.
When that returns, it'll be stored in the local db.

The second call has "wait=True", if it's available in the local db then the results
are returned immediately, otherwise it'll be as before and it'll wait until
the WhatsApp service replies.

### DIFF
--- a/go-app-ussd_clinic.js
+++ b/go-app-ussd_clinic.js
@@ -709,10 +709,15 @@ go.app = function() {
                 });
         };
 
-        self.is_valid_recipient_for_pilot = function (params) {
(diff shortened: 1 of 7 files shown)


smn:
@erikh360 @KaitCrawford @rudigiesler ready for review!

rudigiesler (approved):

PR #132, opened 2017-10-04
Offer eligible users to receive messages over pilot channel

rudigiesler:


go-app-ussd_pmtct_seed.js
```diff
@@ -58,30 +58,109 @@ go.app = function() {
...
-                        "Recently tested HIV positive? You are not alone, many other pregnant " +
-                        "women go through this. Visit b-wise.mobi or call the AIDS Helpline " +
-                        "0800 012 322"
-                    ))
-                );
-            });
+            return self
+                .get_channel()
+                .then(function(channel) {
+                    console.log(channel);
```
alexmuller: Is this log intended to be here or is it for debugging?

go-app-ussd_pmtct_seed.js
```diff
@@ -113,7 +192,19 @@ go.app = function() {
...
+            //          anything with the result.
+            //
+            //          The idea is that the check continues to happen in the background and will be ready
+            //          when we need it. This way we minimise any timeout penalty during the registration.
+            return self
+                .is_valid_recipient_for_pilot({
+                    address: address,
+                    wait: false
+                }).then(function(res) {
+                    return self.states.create("state_check_pmtct_subscription");
```
alexmuller, line 204: Do we need to check the return value of `.is_valid_recipient_for_pilot()` or does that happen automatically somehow?

alexmuller (approved):
I think I understand what's going on here, just a couple of comments for the bits I don't 👍

go-app-ussd_pmtct_seed.js
rudigiesler: No, it was for debugging. I removed it from the source file, but didn't update the built file. https://github.com/praekeltfoundation/ndoh-jsbox/pull/132/commits/dc2782cb23f7ac834074c02c130878e46eac7c4d removes this.

go-app-ussd_pmtct_seed.js
rudigiesler: The return value of `is_valid_recipient_for_pilot` is sent as the argument to the function specified in `.then`, which in this case is `res`.

But in this case we ignore it. This is the `wait: false` call, which just tells the API to do and cache the query, so that when we call it later with `wait: true`, it has hopefully by that point completed and we get a quick response.

If you're talking about checking the response code, then yes, that should go through the errback, which by default should trigger the error state.

alexmuller: Makes sense, thanks! 👍

PR #133, opened 2017-10-26
Add statement to debug 'undefined is not a function' error

dubesoftware:


src/sms_nurse.js
```diff
@@ -115,41 +115,43 @@ go.app = function() {
         self.states.add("states_start", function() {
             var msisdn = utils.normalize_msisdn(self.im.user.addr, "27");
             self.im.user.set_answer("operator_msisdn", msisdn);
+            // Add debug statement
+            self.im.log("Begin running spaghetti...");
```
rudigiesler, line 119: Maybe we can make this a bit more descriptive? Like "Entering states_start", or "Determining identity's target state"

PR #134, opened 2017-10-26
Add debug code to identify source of 'undefined is not a function' error

dubesoftware:


src/sms_nurse.js
```diff
@@ -115,6 +115,8 @@ go.app = function() {
         self.states.add("states_start", function() {
             var msisdn = utils.normalize_msisdn(self.im.user.addr, "27");
             self.im.user.set_answer("operator_msisdn", msisdn);
+            // Add debug statement to identify source of 'undefined is not a function' error
+            self.im.log("\n".join(" Entering states_start ...".join("\n")));
```
rudigiesler: In node (at least the version we use), strings don't have a `.join` method.

I think what you want here, is either a `"\n" + "Entering...`, or just `"\nEntering st..."`

PR #137, opened 2017-11-01
[NUR-426] Offer IP messaging signup for NurseConnect

alexmuller:


.travis.yml
```diff
@@ -1,3 +1,3 @@
 language: node_js
 node_js:
-  - "0.10"
+  - "4"
```
KaitCrawford: Unfortunately, our containers run v 0.10 of Node so we should definitely be testing on that. I'm not aware of any plans to upgrade this soon. You can probably leave the 4 here so that things are tested on both though

alexmuller: Good point, updated so that we test on both but 0.10 is still required. Thanks!

KaitCrawford (approved):
:+1:

PR #138, opened 2017-11-02
[NUR-426] Bugfix lodash and dependencies

alexmuller:
Sorry this diff is a bit boring to review.

## Remove easy use of `_` from `test_utils`

We can do this in plain JavaScript

## Lint test files

To keep them tidy. Includes some duplicate object keys and some unused dependencies.

## Lint undefined variables

This helps us catch places where a dependency is included because of the tests but isn't declared properly in application code.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,6 +1,14 @@
 {
+  "globals": {
+    "beforeEach": true,
+    "describe": true,
+    "go": true,
+    "it": true
+  },
   "laxbreak":  true,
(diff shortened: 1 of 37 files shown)


rudigiesler (approved):
Great! I didn't think of using linting to solve this issue.

I also didn't know that the linter wasn't running on the test files :/

alexmuller:
Thanks, I guess it's better than nothing! :)

PR #141, opened 2017-11-14
Add translations to docker image

erikh360:


Dockerfile
```diff
@@ -4,6 +4,7 @@ MAINTAINER Praekelt Foundation <dev@praekeltfoundation.org>
 # Install nodejs dependencies
 COPY package.json /app/package.json
 COPY config/go-app-ussd_popi*.json /app/
+COPY config/go-app-sms_pmtct*.json /app/
```
KaitCrawford: shouldn't this be `go-app-ussd_pmtct` or did you make changes to the sms stuff as well?

erikh360: Thanks for that, copy and paste fail. Fixed it in latest commit.

KaitCrawford (approved):
:+1:

PR #143, opened 2018-02-15
Update copy on USSD line

rudigiesler:
With us adding the possibility of receiving messages over either WhatsApp or SMS, there are some copy changes that need to be made to ensure that we don't refer to SMS messaging, as this could be confusing for the user if they're signed up for WhatsApp messaging


erikh360 (approved):
Looks good!

PR #146, opened 2018-03-14
Add channel switch keywords

rudigiesler:
Adding keywords in order to switch between the different available channels


erikh360 (approved):
I like the dynamic fixtures! 🍰

PR #147, opened 2018-03-14
Change SMS keyword logic

rudigiesler:
Currently, the logic is that if the FIRST word matches the keyword, then it is triggered.

We want to change this, so that the whole message has to match the keyword (minus non-word characters and whitespace, and case insensitive)

### DIFF
--- a/go-app-sms_inbound.js
+++ b/go-app-sms_inbound.js
@@ -288,7 +288,8 @@ go.app = function() {
                                     return self.states.create("state_opt_in_enter");
                                 case "BABY": case "USANA": case "SANA": case "BABA":
                                 case "BABBY": case "LESEA": case "BBY": case "BABYA":
-                                case "OBABY": case "NGWANA":
+                                case "OBABY": case "NGWANA": case "BABY BOY":
+                                case "BABY GIRL":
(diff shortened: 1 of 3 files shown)


erikh360 (commented):
Just the one tiny comment and a question:

What is the thinking around this? I mean `BABY BOY` would have done the baby change before this code change anyway, this just means other messages like `SMS please` will go to the helpdesk instead of actually switching to sms.

rudigiesler:
@erikh360 Yup, so we did a pull of the inbound messages by looking at the logs, and it turns out that we were switching mother's who were trying to tell us that they had a miscarriage. There will be some cases like `SMS please` that will go through, but looking at the actual messages coming in, that was a very small percentage of messages, most mothers are sending in just the keyword.

So while it will slightly increase the amount of messages going into the helpdesk, it seems like it won't be by much, and it will hopefully reduce the amount of false positives. We'll also evaluate it again at a later stage, and see if there are any more multi-word keywords that we need to add.

erikh360:
@rudigiesler That makes perfect sense.

erikh360 (approved):

PR #152, opened 2018-04-25
Add additional keywords to whatsapp switch

rudigiesler:
We have noticed that some people are replying `"WA" to move over`, so we want to add this as a keyword so that it doesn't go through to the helpdesk

### DIFF
--- a/src/sms_inbound.js
+++ b/src/sms_inbound.js
@@ -149,12 +149,12 @@ go.app = function() {
                                     return self.states.create("state_opt_in_enter");
                                 case "BABY": case "USANA": case "SANA": case "BABA":
                                 case "BABBY": case "LESEA": case "BBY": case "BABYA":
-                                case "OBABY": case "NGWANA": case "BABY BOY":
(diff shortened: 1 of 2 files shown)


src/sms_inbound.js
```diff
@@ -149,12 +149,12 @@ go.app = function() {
                                     return self.states.create("state_opt_in_enter");
                                 case "BABY": case "USANA": case "SANA": case "BABA":
                                 case "BABBY": case "LESEA": case "BBY": case "BABYA":
-                                case "OBABY": case "NGWANA": case "BABY BOY":
-                                case "BABY GIRL":
+                                case "OBABY": case "NGWANA": case "BABYBOY":
+                                case "BABYGIRL":
```
rudigiesler, line 153: I noticed that these keywords have spaces in them, but in our processing of the inbound message, we remove all non-word characters, which include spaces. So we should be comparing the message to keywords without spaces.

erikh360 (approved):
When travis is happy!

🍰 for the other space issue!

PR #153, opened 2018-05-11
remove annotation API calls for wassup, this stuff is tracked elsewhere now

smn:

PR #154, opened 2018-05-14
Switch over to new /api/v1/lookups/ API

smn:
The /api/v1/numbers/check API is being deprecated.
This has PR #153 merged in so it would help to
review that one first.

### DIFF
--- a/src/ussd_public.js
+++ b/src/ussd_public.js
@@ -274,29 +274,6 @@ go.app = function() {
             return registration_info;
         };
 
-        self.annotate_pilot = function (metadata) {
-            var pilot_config = self.im.config.pilot || {};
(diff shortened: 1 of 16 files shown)


src/ussd_public.js
```diff
@@ -621,9 +598,9 @@ go.app = function() {
             var api_token = pilot_config.api_token;
             var api_number = pilot_config.api_number;
 
-            var default_params = _.merge({
-                number: api_number,
-            }, params);
+            // var default_params = _.merge({
+            //     number: api_number,
+            // }, params);
```
rudigiesler: Remove instead of commenting out?

rudigiesler (commented):
Just one small comment from me, looks good otherwise

PR #155, opened 2018-05-14
Add docker-compose setup for trying out the applications

rudigiesler:
This docker compose setup should allow people to very easily spin up all the required infrastructure to be able to play around with the USSD applications.

Currently, only the Public USSD line is implemented.


docker-compose/docker-compose.yml
```diff
@@ -0,0 +1,82 @@
...
+    environment:
+      - POSTGRES_PASSWORD=postgres
+    restart: always
+
+  identitystore:
+    image: praekeltfoundation/seed-identity-store
+    links:
+      - postgres
+    environment:
+      - IDENTITIES_DATABASE=postgres://postgres:postgres@postgres/postgres
```
erikh360: I see all the seed services are connecting to a database called postgres, is this intentional?

erikh360 (commented):
This is awesome, just the one question.

docker-compose/docker-compose.yml
rudigiesler: Kind of. We could change this, but this means that I just need to create a token for one application. Let me take a look to see how difficult it is to have more databases in the postgres docker image

Boxroma:
This is Great thanks!!

rudigiesler:
@erikh360 Updated with separate DBs. Ready for review again

erikh360 (approved):

PR #156, opened 2018-05-22
Feature/support for number change

OMosimege:
added support for number change when the user does not have access to old sim card

### DIFF
--- a/go-app-ussd_popi_user_data.js
+++ b/go-app-ussd_popi_user_data.js
@@ -716,9 +716,14 @@ go.app = function() {
             next: "state_start"
           });
         });
+
+        self.add('state_old_number', function(name){
+          return new FreeText(name, {
+            question: $("Please enter the number you receive MomConnect messages on."),
+            next: 'state_find_identity'
+          })
+        })
     });
-
-
     return {
         GoNDOH: GoNDOH
     };
--- a/go-app-ussd_popi_user_data.js
+++ b/go-app-ussd_popi_user_data.js
@@ -7,6 +7,7 @@ go.app = function() {
     var MetricsHelper = require('go-jsbox-metrics-helper');
     var Q = require('q');
     var _ = require('lodash');
+    var moment = require('moment');
     var App = vumigo.App;
     var EndState = vumigo.states.EndState;
     var ChoiceState = vumigo.states.ChoiceState;
@@ -762,7 +763,7 @@ go.app = function() {
           return new FreeText(name, {
             question: $("Thank you. To change your mobile number we first need to " +
                       "verify your identity. Please enter your SA ID number now."),
-            next: 'state_get_langauge'
+            next: 'state_get_language'
           });
         });
 
@@ -770,7 +771,7 @@ go.app = function() {
           return new FreeText(name, {
             question: $("Thank you. To change your mobile number we first need to " +
                       "verify your identity. Please enter your passport number now."),
-            next: 'state_get_langauge'
+            next: 'state_get_language'
           });
         });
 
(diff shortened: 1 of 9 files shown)


src/ussd_popi_user_data.js
```diff
@@ -691,14 +692,268 @@ go.app = function() {
...
+            question: $("Thank you. To change your mobile number we first need to " +
+                      "verify your identity. Please enter your passport number now."),
+            next: 'state_get_language'
+          });
+        });
+
+        self.add('state_get_date_of_birth', function(name){
+          return new FreeText(name, {
+            question: $("Thank you. To change your mobile number we first need to " +
+                      "verify your identity. Please enter your date of birth in the following format: dd/mm/yyyy"),
```
erikh360: How easy is it to enter this format including the `/` on a feature phone?

src/ussd_popi_user_data.js
```diff
@@ -691,14 +692,268 @@ go.app = function() {
...
+          if (identity.details.passport_no !== passport_no){
+            return self.states.create('state_incorrect_security_answers');
+          }
+        }
+        if(date_of_birth !== undefined){
+          if (identity.details.mom_dob !== date_of_birth){
+            return self.states.create('state_incorrect_security_answers');
+          }
+        }
+      return self.states.create('state_enter_new_phone_number');
```
erikh360: This needs to be indented.

src/ussd_popi_user_data.js
```diff
@@ -691,14 +692,268 @@ go.app = function() {
...
+        var msisdn = utils.normalize_msisdn(self.im.user.get_answer('state_enter_new_phone_number'), "27");
+        var channel = self.im.user.get_answer('state_new_number_channel');
+        return hub.create_change({
+          "registrant_id": self.im.user.get_answer('user_identity').id,
+          "action": "momconnect_change_msisdn",
+          "data": {
+                    "msisdn": msisdn
+                  }
+        }).then(function(){
+          return hub.create_change({
```
erikh360: Looks like we will always create a `switch_channel `, even if the user stays on the same channel.

test/ussd_popi_user_data.test.js
```diff
@@ -120,6 +121,428 @@ describe('app', function() {
...
+                .check.interaction({
+                  state: 'state_get_sa_id',
+                  reply: "Thank you. To change your mobile number we first " +
+                          "need to verify your identity. Please enter your SA ID number now."
+                })
+                .run();
+              });
+
+              describe('after user enters ID number', function(){
+                it('should lits the language options', function(){
```
erikh360: Spelling. lits -> list

test/ussd_popi_user_data.test.js
```diff
@@ -120,6 +121,428 @@ describe('app', function() {
...
+                      reply:
+                        'Sorry that is not the correct format. Please enter your date of ' +
+                        'birth in the following format: dd/mm/yyyy. For example 19/05/1990'
+                    })
+                    .run();
+                });
+              });
+
+              describe('after user enters date of birth', function(){
+                it('should lits the language options', function(){
```
erikh360: Same here for the spelling.

test/ussd_popi_user_data.test.js
```diff
@@ -120,6 +121,428 @@ describe('app', function() {
...
+                .check.interaction({
+                  state: 'state_get_passport_no',
+                  reply: "Thank you. To change your mobile number we first " +
+                          "need to verify your identity. Please enter your passport number now."
+                })
+                .run();
+              });
+
+              describe('after user enters passport number', function(){
+                it('should lits the language options', function(){
```
erikh360: Another spelling.

go-app-ussd_popi_user_data.js
```diff
@@ -694,14 +695,268 @@ go.app = function() {
...
+          choices: [
+              new Choice(
+                  'state_enter_new_phone_number',
+                  $("Try again")),
+              new Choice('state_exit', $("Exit"))
+          ]
+        });
+      });
+
+      self.add('state_new_number_channel', function(name) {
```
erikh360, line 912: I see we don't check if the new number is on whatsapp, is the thinking that we just switch them to SMS if the first message fails?

erikh360 (commented):
Looks good. I have left a few minor comments and some questions.
🍰

go-app-ussd_popi_user_data.js
rudigiesler: Yes that is the thinking. Now that we have the automatic switch in place, I think it's better to rely on that, rather than rely on the number check API, which has rate-limiting.

src/ussd_popi_user_data.js
rudigiesler: As it turns out, after doing some testing, it's actually impossible to enter the `/` on a feature phone on a USSD line, we'll be changing this to the `*` character, which is much easier to enter.

erikh360 (approved):
Nice!

PR #157, opened 2018-05-23
Remove need for double-opt in

smn:
Currently we're requiring two opt-ins which is fairly counter intuitive and unnecessary.
The first opt in is whether they want to receive MomConnect messages, the second opt-in is whether they want to receive via SMS or WhatsApp.

I think we should collapse it into a single opt-in for MomConnect.
If the mother is WhatsAppable then default to WA, otherwise SMS.

The welcome message should make it very explicit on how the mother can change to WA or SMS.

### DIFF
--- a/go-app-ussd_clinic.js
+++ b/go-app-ussd_clinic.js
@@ -573,8 +573,8 @@ go.app = function() {
             return new ChoiceState(name, {
                 question: $(
                     'We need to collect, store & use her info. She ' +
-                    'may get messages on public holidays & weekends. ' +
-                    'Does she consent?'),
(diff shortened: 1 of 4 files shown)


rudigiesler (commented):
@smn We'll need to run `grunt build` or `npm test` to regenerate the app file. Otherwise looks good to me, 👍

smn:
@rudigiesler yeah added the build now. I've also removed the randomisation stuff, simplifies the state machine slightly.

rudigiesler:
https://github.com/praekeltfoundation/ndoh-jsbox/pull/188 achieves the same goal, closing this PR

PR #164, opened 2018-06-27
Feature/ussd nurse rapidpro

OMosimege:


src/ussd_nurse_rapidpro.js
```diff
@@ -0,0 +1,116 @@
...
+    var GoNDOH = App.extend(function(self) {
+        App.call(self, 'state_start');
+        var utils = SeedJsboxUtils.utils;
+
+        //variables for services
+        var is;
+
+        self.init = function() {
+            // initialise services
+            is = new IdentityStore(
```
erikh360, line 21: I'm guessing we're probably not going to use this. Same for `IdentityStore` variable declared above.

src/ussd_nurse_rapidpro.js
```diff
@@ -0,0 +1,116 @@
...
+            return new ChoiceState(name, {
+                question: ("To register, your friend's info needs to be collected, stored and used. "+
+                            "They may receive messages on public holidays. Do they agree?"),
+                choices: [
+                    new Choice('yes', ('Yes')),
+                    new Choice('no', ('No')),
+                ],
+                next: function(choice) {
+                    if (choice.value === 'yes') {
+                        return 'state_agree_to_register';
```
erikh360: I see `state_weekly_messages` and `state_friend_register` both go to `state_agree_to_register` when the user agrees to share information, we will need to set a variable so we know who is registering when we get to `state_agree_to_register`.

test/ussd_nurse_rapidpro.test.js
```diff
@@ -0,0 +1,182 @@
...
+            tester = new AppTester(app);
+
+            tester
+                .setup.config.app({
+                    name: 'ussd_nurse_rapidpro',
+                    env: 'test',
+                    no_timeout_redirects: ['state_start'],
+                    channel: '*134*550*5#',
+                    services: {
+                        identity_store: {
```
erikh360, line 24: This is not necessary.

test/ussd_nurse_rapidpro.test.js
```diff
@@ -0,0 +1,182 @@
...
+                            url: 'http://is/api/v1/',
+                            token: 'test IdentityStore'
+                        }
+                    },
+                })
+                .setup(function(api) {
+                    // add fixtures for services used
+                    fixtures_IdentityStore().forEach(api.http.fixtures.add); // 120 ->
```
erikh360, line 32: These fixtures mock the identity store so will probably be replaced with the companion app.

erikh360 (commented):
Looks good so far! 🍰 

Few comments but mostly related to the identity store(seed) being replaced by the companion app. If you want to keep it there to replace later that is also fine, then you can ignore those comments.

Two things I would suggest for future PRs:
Add a description of what the PR contains.
When creating the branch include the Jira ticket number, for example `MOMZA-111-ussd-nurse-rapidpro-app`, it then shows all the branches and PRs on the ticket.

src/ussd_nurse_rapidpro.js
OMosimege: yes, will remove that part after the companion app is complete, I just didn't know how to test without using the identity store. If you do have suggestions I'm open for help :)

src/ussd_nurse_rapidpro.js
OMosimege: Ahh I see what you mean. But according to the flow diagram I was given, they both branch in and follow the same flow thereafter, which would make it hard if I differentiate them into 2 different variables. Or is it still wise to do so regardless?

test/ussd_nurse_rapidpro.test.js
OMosimege: agreed, will be sure to note it out and have it replaced at a later stage thanks

test/ussd_nurse_rapidpro.test.js
OMosimege: yes, they will be

src/ussd_nurse_rapidpro.js
erikh360: It's fine that they go to the same state but when you get there you might want to know who is registering, the person dialling or a friend. So you could do something like:
`self.im.user.set_answer("friend_registration", true);`

erikh360 (approved):

PR #167, opened 2018-07-05
Feature/ussd nurse rapidpro

OMosimege:
Nurse Connect is moving over to rapid pro

This PR include an additional menu state for subscribed users, tests for that menu state, added comments for test states, and temporary fixtures that were added (temporary as they will be replaced by functions that will replace Seed functions)

Comments are also added for lines that will be later replaced

### DIFF
--- a/go-app-ussd_nurse_rapidpro.js
+++ b/go-app-ussd_nurse_rapidpro.js
@@ -5,6 +5,7 @@ go.app = function() {
     var vumigo = require('vumigo_v02');
     var MenuState = vumigo.states.MenuState;
     var ChoiceState = vumigo.states.ChoiceState;
+    var PaginatedChoiceState = vumigo.states.PaginatedChoiceState;
     var FreeText = vumigo.states.FreeText;
     var Choice = vumigo.states.Choice;
     var SeedJsboxUtils = require('seed-jsbox-utils');
@@ -43,7 +44,7 @@ go.app = function() {
                 else state_not_registered
             */
             if(msisdn === "+27820001003"){ 
-                return self.states.create('state_registered_1');
+                return self.states.create('state_registered');
             } else {
                 return self.states.create('state_not_registered');
             }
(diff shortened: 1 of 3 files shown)


go-app-ussd_nurse_rapidpro.js
```diff
@@ -36,24 +37,54 @@ go.app = function() {
...
-                next: 'state_nurse_connect_options'
+                next: 'state_not_registered_menu'
+            });
+        });
+
+        self.states.add('state_registered_1', function(name) {
+            return new MenuState(name, {
+                question: ("Welcome back to NurseConnect. Do you want to:"),
+                choices: [
+                    new Choice('state_friend_registration', ('Help a friend sign up')),
```
sewagodimo: people can sign their friends up for nurse connect? Shuu

OMosimege: not friend, but a fellow nurse "friend". The state option is called `Help a friend register`. Do you find it better that the state be named differently?

sewagodimo: colleague maybe?
Friend sounds like ichommie zakho

test/ussd_nurse_rapidpro.test.js
```diff
@@ -51,12 +51,75 @@ describe('app', function() {
...
+                    return tester
+                    .setup.user.addr('27820001003')
+                    .inputs(
+                        {session_event: 'new'}  // dial in
+                    )
+                    .check.interaction({
+                        state: 'state_registered_1',
+                        reply: [
+                                "Welcome back to NurseConnect. Do you want to:",
+                                "1. Help a friend sign up",
```
sewagodimo: or maybe `fellow nurse` just to be clear

sewagodimo (approved):
I don't really think the `friend` thing is a big deal so..
:cake:

test/ussd_nurse_rapidpro.test.js
OMosimege: The SxD defined the words in the flow. I cant change it, might clash with the content team

go-app-ussd_nurse_rapidpro.js
OMosimege: I'll rename it to "other_registration". See comment bellow

src/ussd_nurse_rapidpro.js
```diff
@@ -33,24 +34,54 @@ go.app = function() {
...
         self.states.add('state_not_registered', function(name){
             return new FreeText(name, {
                 question: ("Welcome to NurseConnect, where you can stay up to date with "+
                 "maternal & child health. Reply '1' to start."),
-                next: 'state_nurse_connect_options'
+                next: 'state_not_registered_menu'
+            });
+        });
+
+        self.states.add('state_registered_1', function(name) {
```
KaitCrawford: I wanted to mention yesterday that I think it would be better to combine this state and the one below it since the jsbox is designed to handle splitting stuff like that. But that is simply a suggestion.

OMosimege: could you please help me out with this? I do want to join them somehow and make the code neater

KaitCrawford: @OMosimege I think you should be able to put all the choices in one state and the jsbox utils should automatically handle splitting it. You might not even have to change your tests

OMosimege: ahh I see. But I see here that the jsbox utils are part of Seed, in which for this specific project all Seed functions are going to be replaced, so I'll put a note in it and leave it as is for now

KaitCrawford: @OMosimege sorry, I used the wrong term there. Since the ChoiceState is part of the vumigo library I expect that is where it would be handled.

PR #168, opened 2018-07-09
Send identity optout to hub

erikh360:
Part of this change: `https://github.com/praekeltfoundation/ndoh-hub/pull/160`.

Identity optout needs to go through the hub to insure that we send it to jembi before we remove the msisdn from the identity.

### DIFF
--- a/go-app-ussd_popi_user_data.js
+++ b/go-app-ussd_popi_user_data.js
@@ -678,32 +678,30 @@ go.app = function() {
         });
 
         self.add('state_optout', function(name) {
+            var optout_info = {
+                "optout_type": "forget",
+                "identity": self.im.user.answers.operator.id,
+                "reason": "unknown",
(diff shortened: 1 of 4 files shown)


DevChima (commented):
I'm not very clear on this but I'm going to +1.

erikh360:
@DevChima Which part? Maybe I can clarify it for you.

DevChima:
@erikh360  This part sort of throws me off. 
https://github.com/praekeltfoundation/ndoh-jsbox/pull/168/files#diff-aaba838434d1a8ee094055239f97ba3dR690
Why is it returned.

erikh360:
@DevChima each state needs to return the next state that it needs to move to. You will see it does a `.then` that returns the next state.

DevChima:
@erikh360 :+1:

PR #170, opened 2018-07-17
Feature/ussd nurse rapidpro

OMosimege:
added flow for if registered user chooses to opt out, and tests for that


erikh360 (approved):
Looks good!

PR #171, opened 2018-07-18
Feature/ussd nurse rapidpro

OMosimege:
Removed identity store
Added function normalising msisdns
Added background WhatsApp check
Added facility code validation function, whatsapp check, and jembi api call function

### DIFF
--- a/go-app-ussd_nurse_rapidpro.js
+++ b/go-app-ussd_nurse_rapidpro.js
@@ -2,7 +2,9 @@ var go = {};
 go;
 
 go.app = function() {
+    var _ = require('lodash');
     var vumigo = require('vumigo_v02');
+    var Q = require('q');
     var MenuState = vumigo.states.MenuState;
     var ChoiceState = vumigo.states.ChoiceState;
     var PaginatedChoiceState = vumigo.states.PaginatedChoiceState;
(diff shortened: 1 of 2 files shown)


go-app-ussd_nurse_rapidpro.js
```diff
@@ -2,33 +2,110 @@ var go = {};
...
+    var utils = SeedJsboxUtils.utils;
     var App = vumigo.App;
 
     var GoNDOH = App.extend(function(self) {
         App.call(self, 'state_start');
-        var utils = SeedJsboxUtils.utils; //replace with utils not in Seed
 
         //variables for services
-        var is;
+        /*var is;*/
```
erikh360, line 23: Can we remove this instead of commenting it out?

go-app-ussd_nurse_rapidpro.js
```diff
@@ -2,33 +2,110 @@ var go = {};
...
 
         self.init = function() {
             // initialise services
             //replace identity store
-            is = new IdentityStore(
+            /*is = new IdentityStore(
```
erikh360, line 28: Can we remove this instead of commenting it out?

erikh360 (commented):
Just the two comments. Nice!

go-app-ussd_nurse_rapidpro.js
OMosimege: Oh thanks for pointing this out

go-app-ussd_nurse_rapidpro.js
OMosimege: ahh this too thanks

erikh360 (approved):

PR #172, opened 2018-07-19
Feature/ussd nurse rapidpro

OMosimege:
Added a state_facname, added details in state facade that lead to state_facname

Added fixtures for WhatsApp and jembi, and added tests for state_facade that lead to state_facname or loop back to state_faccode

### DIFF
--- a/test/ussd_nurse_rapidpro.test.js
+++ b/test/ussd_nurse_rapidpro.test.js
@@ -19,8 +19,6 @@ describe('app', function() {
                 .setup.config.app({
                     name: 'ussd_nurse_rapidpro',
                     env: 'test',
-                    testing_today: 'July 1, 2018 07:07:07',
-                    testing_message_id: '0170b7bb-978e-4b8a-35d2-662af5b6daee',
                     no_timeout_redirects: ['state_start'],
                     channel: '*134*550*5#',
                     jembi: {
--- a/test/ussd_nurse_rapidpro.test.js
+++ b/test/ussd_nurse_rapidpro.test.js
@@ -2,7 +2,9 @@ var vumigo = require('vumigo_v02');
 var AppTester = vumigo.AppTester;
 var fixtures_IdentityStore = require('./fixtures_identity_store');
 var fixtures_Pilot = require('./fixtures_pilot');
+var fixtures_Jembi = require('./fixtures_jembi');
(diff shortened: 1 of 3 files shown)


test/ussd_nurse_rapidpro.test.js
```diff
@@ -17,8 +19,20 @@ describe('app', function() {
                 .setup.config.app({
                     name: 'ussd_nurse_rapidpro',
                     env: 'test',
+                    testing_today: 'July 1, 2018 07:07:07',
+                    testing_message_id: '0170b7bb-978e-4b8a-35d2-662af5b6daee',
```
erikh360: I don't see where this is used?

test/ussd_nurse_rapidpro.test.js
```diff
@@ -439,7 +454,90 @@ describe('app', function() {
...
+                            .inputs(
+                                {session_event: 'new'}
+                                ,'1' //chooses to start nurse connect
+                                ,'1'  // chooses to sign up
+                                ,'1' //chooses yes to subscribe
+                                ,'0820001004'  // state_msisdn
+                                ,'1' //chooses yes to opt in
+                                ,'12345' //enters facility code
+                            )
+                            .run();
```
erikh360: We're not checking here if it actually goes to the `state_faccname` state?

erikh360 (commented):
Just a few comments.

test/ussd_nurse_rapidpro.test.js
OMosimege: Thanks for pointing this out, need to add more to make sure it goes to the correct state

test/ussd_nurse_rapidpro.test.js
OMosimege: hmm, true. Don't need it

rudigiesler: Does this still need to be removed?

rudigiesler (commented):
Just one small comment. Looks good otherwise

test/ussd_nurse_rapidpro.test.js
OMosimege: Thanks, forgot to remove it

rudigiesler (approved):

PR #173, opened 2018-07-27
Add contact check on rapidpro for nurseconnect

rudigiesler:
This PR adds a check with rapidpro for whether the contact is recognised or not, as well as some test cleanup


OMosimege (approved):
Looks great! 🍰

PR #175, opened 2018-07-31
Feature/nur 469 add user detail change flows

OMosimege:
Added flows for when user changes details.
includes states to change facility code, to change ID/ Passport number, and to change sanc no

### DIFF
--- a/src/ussd_nurse_rapidpro.js
+++ b/src/ussd_nurse_rapidpro.js
@@ -1,6 +1,5 @@
 go.app = function() {
     var _ = require('lodash');
-    var moment = require('moment');
     var vumigo = require('vumigo_v02');
     var Q = require('q');
     var MenuState = vumigo.states.MenuState;
@@ -410,9 +409,9 @@ go.app = function() {
                     } else {
                         /*
                             add function to update info against operator id
+                            send this update to RapidPro
                             then lead to state where user details are successfully changed
                         */
-                        self.im.user.set_answer("sanc_reg_no", content);
                         return null;
                     }
                 },
@@ -458,10 +457,6 @@ go.app = function() {
                         add function to update date of birth info against operator id
                         then lead to state where user details are successfully changed
                         */
-                    var date_of_birth = utils.extract_za_id_dob(content);
-                    self.im.user.set_answer("id_type", "sa_id");
-                    self.im.user.set_answer("sa_id_no", content);
-                    self.im.user.set_answer("dob", date_of_birth);
                     }
                 },
                 next: 'state_end_detail_changed',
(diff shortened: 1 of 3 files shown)


src/ussd_nurse_rapidpro.js
```diff
@@ -370,7 +371,161 @@ go.app = function() {
...
+                check: function(content) {
+                    if (!utils.check_valid_number(content)
+                        || content.length !== 8) {
+                        return error;
+                    } else {
+                        /*
+                            add function to update info against operator id
+                            then lead to state where user details are successfully changed
+                        */
+                        self.im.user.set_answer("sanc_reg_no", content);
```
rudigiesler: This will already be saved under `state_change_sanc`, so we don't need to save it here again.

src/ussd_nurse_rapidpro.js
```diff
@@ -370,7 +371,161 @@ go.app = function() {
...
+                        /*
+                        add function to update id type info against operator id
+                        add function to update id number info against operator id
+                        add function to update date of birth info against operator id
+                        then lead to state where user details are successfully changed
+                        */
+                    var date_of_birth = utils.extract_za_id_dob(content);
+                    self.im.user.set_answer("id_type", "sa_id");
+                    self.im.user.set_answer("sa_id_no", content);
+                    self.im.user.set_answer("dob", date_of_birth);
```
rudigiesler: The `id_type` will be saved under `state_change_id_no`, and the rest under `state_id_no`, so we don't need to save them again.

src/ussd_nurse_rapidpro.js
```diff
@@ -370,7 +371,161 @@ go.app = function() {
...
+                            add function to update id type info against operator id
+                            add function to update id number info against operator id
+                            add function to update date of birth info against operator id
+                            then lead to state where user details are successfully changed
+                        */
+                       var date_of_birth = moment(content, 'DDMMYYYY').format('YYYY-MM-DD');
+                        self.im.user.set_answer("id_type", "passport");
+                        self.im.user.set_answer("passport_no", self.im.user.answers.state_passport_no);
+                        self.im.user.set_answer("passport_origin", self.im.user.answers.state_passport);
+                        self.im.user.set_answer("dob", date_of_birth);
```
rudigiesler: Same here with these answers, we don't need to save them since we already have them automatically saved.

rudigiesler (commented):
Just a few comments about saving details that are already automatically saved. Otherwise looks good :)

src/ussd_nurse_rapidpro.js
OMosimege: ahh I see, will remove them. Should I remove the comments as well?

src/ussd_nurse_rapidpro.js
OMosimege: should I remove setting the date of birth as well? (and should the comments also be removed?)

src/ussd_nurse_rapidpro.js
OMosimege: same question as above

src/ussd_nurse_rapidpro.js
rudigiesler: Yes, I think we can do the date of birth generation when we send the update to RapidPro, instead of here.

src/ussd_nurse_rapidpro.js
rudigiesler: I think the comments are fine. We need some reminder here that we need to send this update to RapidPro

OMosimege: okay thanks

src/ussd_nurse_rapidpro.js
OMosimege: noted

rudigiesler (approved):

PR #177, opened 2018-07-31
Feature/nur 469 add change phone and persal no

OMosimege:
Added state to change nurse persal no. and subsequent tests for that

Added state to change user phone number, and a permission denied end state

Description of updates of user-detail changes that will be made to rapid pro have been put in comment blocks

### DIFF
--- a/go-app-ussd_nurse_rapidpro.js
+++ b/go-app-ussd_nurse_rapidpro.js
@@ -712,6 +712,75 @@ go.app = function() {
             });
         });
 
+        self.states.add('state_change_number', function(name) {
+            var question = $("Please enter the new number on which you want to receive messages, e.g. 0736252020:");
+            var error = $("Sorry, the format of the mobile number is not correct. Please enter the new number on which you want to receive messages, e.g. 0736252020");
+            return new FreeText(name, {
+                question: question,
+                check: function(content) {
+                    if (!utils.is_valid_msisdn(content, 0, 10)) {
+                        return error;
+                    }
+                },
+                next: function(content) {
+                    return 'state_check_optout_change';
+                }
+            });
+        });
+
+        self.states.add('state_check_optout_change', function(name) {
+            var new_msisdn = utils.normalize_msisdn(self.im.user.answers.state_change_number, '27');
+            self.im.user.set_answer("new_msisdn", new_msisdn);
+
+            // If the contact exists and is in the opt out group, then it's opted out, otherwise not
+            var contact = self.im.user.answers.registrant_contact;
+            if(contact && self.is_contact_in_group(contact, 'opted-out')){
+                return self.states.create('state_opt_in_change');
+            }
+            return self.states.create('state_switch_new_nr');
+        });
+
+        self.states.add('state_opt_in_change', function(name) {
+            return new ChoiceState(name, {
+                question: $("This number opted out of NurseConnect messages before. Please confirm that you want to receive messages again on this number?"),
+                choices: [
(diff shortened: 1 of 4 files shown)


go-app-ussd_nurse_rapidpro.js
```diff
@@ -690,6 +712,75 @@ go.app = function() {
...
+        self.states.add('state_check_optout_change', function(name) {
+            var new_msisdn = utils.normalize_msisdn(self.im.user.answers.state_change_number, '27');
+            self.im.user.set_answer("new_msisdn", new_msisdn);
+
+            // If the contact exists and is in the opt out group, then it's opted out, otherwise not
+            var contact = self.im.user.answers.registrant_contact;
+            if(contact && self.is_contact_in_group(contact, 'opted-out')){
+                return self.states.create('state_opt_in_change');
+            }
+            return self.states.create('state_switch_new_nr');
```
rudigiesler: I think we need to do another check for if the contact is currently subscribed, then we give them an error if they are, because they can't switch to a number that's already receiving messages.

go-app-ussd_nurse_rapidpro.js
```diff
@@ -690,6 +712,75 @@ go.app = function() {
...
+                }
+            });
+        });
+
+        self.states.add('state_check_optout_change', function(name) {
+            var new_msisdn = utils.normalize_msisdn(self.im.user.answers.state_change_number, '27');
+            self.im.user.set_answer("new_msisdn", new_msisdn);
+
+            // If the contact exists and is in the opt out group, then it's opted out, otherwise not
+            var contact = self.im.user.answers.registrant_contact;
```
rudigiesler: I don't think that this is set by anything. This is only set during the registration flow.

go-app-ussd_nurse_rapidpro.js
```diff
@@ -690,6 +712,75 @@ go.app = function() {
             });
         });
 
+        self.states.add('state_change_number', function(name) {
+            var question = $("Please enter the new number on which you want to receive messages, e.g. 0736252020:");
```
rudigiesler: You can reach this state both from `state_registered` and `state_not_registered_menu`, and in the case of `state_not_registered_menu`, we don't actually know what the number that they want to switch from is, so we need to ask them that somewhere.

go-app-ussd_nurse_rapidpro.js
```diff
@@ -690,6 +712,75 @@ go.app = function() {
...
+                ],
+                next: function(choice) {
+                    if (choice.value === 'yes') {
+                        /*
+                            update this contact will be opted in again
+                         */
+                        return 'state_switch_new_nr';
+                    } else {
+                        return 'state_permission_denied';
+                    }
```
rudigiesler: I think we can change this to a `MenuState`, since all it does it go to another state depending on the user's choice.

test/ussd_nurse_rapidpro.test.js
```diff
@@ -875,6 +875,156 @@ describe('app', function() {
...
+                                {session_event: 'new'}  // dial in
+                                , '2'  // state_subscribed - change num
+                                , '0820001001'  // state_change_num
+                            )
+                            .check.interaction({
+                                state: 'state_end_detail_changed',
+                            })
+                            .run();
+                    });
+                });
```
rudigiesler: We will also need a test to check if the desired number has been opted out before, then they get the yes/no optin screen.

erikh360 (approved):
Nice!

PR #178, opened 2018-07-31
Send detail updates to RapidPro

rudigiesler:
This PR adds the sending of SANC number, Facility code, and identification updates to RapidPro to update the contact details.

It also adds the sending of opt out reasons + running the opt out flow to RapidPro.


OMosimege (approved):
Looks great! 🍰Will follow the trend of the changes

PR #179, opened 2018-08-03
Fix submit NurseConnect registration to OpenHIM

rudigiesler:
For the OpenHIM API, they accept JSON in the body of a POST request, but they don't always return JSON in the body of the response.

This means that we cannot use the JsonApi, because it tries to do a JSON decode on the body. So we have to use the normal HttpApi, and manually encode and decode JSON where necessary.

### DIFF
--- a/go-app-ussd_nurse_rapidpro.js
+++ b/go-app-ussd_nurse_rapidpro.js
@@ -102,10 +102,11 @@ go.OpenHIM = function() {
     var utils = SeedJsboxUtils.utils;
     var moment = require('moment');
 
-    var OpenHIM = Eventable.extend(function(self, json_api, base_url, username, password) {
-        self.json_api = json_api;
+    var OpenHIM = Eventable.extend(function(self, http_api, base_url, username, password) {
+        self.http_api = http_api;
         self.base_url = base_url;
-        self.json_api.defaults.auth = {username: username, password: password};
+        self.http_api.defaults.auth = {username: username, password: password};
+        self.http_api.defaults.headers['Content-Type'] = ['application/json; charset=utf-8'];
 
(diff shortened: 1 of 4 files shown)


go-app-ussd_nurse_rapidpro.js
```diff
@@ -162,6 +164,7 @@ go.app = function() {
     var Choice = vumigo.states.Choice;
     var SeedJsboxUtils = require('seed-jsbox-utils');
     var JsonApi = vumigo.http.api.JsonApi;
+    var HttpApi = vumigo.http.api.HttpApi;
```
OMosimege, line 167: What's the difference between a JSON Api and an HTTP Api? and what caused you to switch to the other?

OMosimege (approved):
Looks great! 🍰

go-app-ussd_nurse_rapidpro.js
rudigiesler: The JSON API automatically takes what you put in `data` (the body of the requests), and encodes it to JSON, and then automatically takes the body of the response, and decodes it expecting it to be JSON

The HTTP API doesn't do that, it just takes whatever string you give it and puts it in the body, and then takes whatever the response is and gives it to you as a string. It doesn't do any of the JSON encoding/decoding for you, you have to do it yourself.

We need this because one of the endpoint that we make a request to doesn't return a JSON reply, it just returns the text `"Accepted"`, which the JSON API tries to decode as JSON, which it fails to do because it's not JSON, and so it breaks and returns an error. Switching to the  HTTP API allows us to control when we should decode it as JSON, and when we should just leave it as a string.

OMosimege: Thanks Rudi

PR #180, opened 2018-08-10
Updates for the NurseConnect RapidPro USSD line

rudigiesler:
There were multiple versions of the flow diagram floating around, and so some of the latest copy changes weren't included when this app was created.

These changes represent the merging of these versions into one final version.


erikh360 (approved):
Nice!

PR #181, opened 2018-08-30
Add support for multiple urn types for rapidpro

rudigiesler:
For NurseConnect RapidPro, we want to support both the tel and whatsapp URN types, and ensure that everything supports multiple urns being returned by Rapidpro


OMosimege (approved):
Looks great 👍

PR #182, opened 2018-09-05
Use engage api for number lookups

erikh360:
I changed the api call to engage and also use the `wa_id` we get from there when creating the contact on Rapidpro.

### DIFF
--- a/go-app-ussd_nurse_rapidpro.js
+++ b/go-app-ussd_nurse_rapidpro.js
@@ -200,27 +200,30 @@ go.app = function() {
         };
 
        self.is_whatsapp_user = function(msisdn, wait_for_response) {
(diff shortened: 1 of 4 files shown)


rudigiesler (commented):
I think we need to do a little more work here with auth.

The way auth in Engage works is, you get a username and password. You can then send that username and password to an endpoint, which gives you a token. That token has an expiry date, however, and so you need to continuously get new tokens as the old ones expire.

So we can't really have the token in the config here, else we'll have to get a token and update the config whenever the token expires (I think it's currently around a week)

erikh360:
@rudigiesler I was thinking of doing that in a separate PR.

rudigiesler (approved):

PR #183, opened 2018-09-05
Get whatsapp token before number check

erikh360:


go-app-ussd_nurse_rapidpro.js
```diff
@@ -199,35 +199,45 @@ go.app = function() {
...
 
-            var params = {
-                "contacts": [msisdn],
-                "blocking": wait_for_response ? "wait" : "no_wait"
-            };
-
-            return new JsonApi(self.im, {
-                headers: {
-                    'Authorization': ['Bearer ' + api_token]
+            return new HttpApi(self.im, {
```
rudigiesler: Why not use a JsonApi here too?

erikh360: Not sure why I ended up with the HttpApi, I thought it only added the header but I see now it decodes the result, then I don't have to do it manually. Thanks! 

Commit incoming.

rudigiesler (approved):

PR #184, opened 2018-09-10
Only send whatsapp urn if available

erikh360:
Fix for RP error: `Invalid URN: whatsapp:undefined. Ensure phone numbers contain country codes.`


rudigiesler (commented):
The reason I had this here is that if they ever install WhatsApp in the future, and request to be switched, then we would have both numbers there in RapidPro, because I don't think that we can add it inside a flow.

Also, I think there are other places where we set the URNs, I think it's where we update the MSISDN

erikh360:
@rudigiesler I've updated it to send the `wa_id` if we have one or the the one we generate from the msisdn.

rudigiesler (approved):

PR #187, opened 2018-09-18
Feature/momza 1029 wa default reg

OMosimege:
This PR is implementing the whatsapp default registration for the CHW USSD line


go-app-ussd_chw.js
```diff
@@ -379,23 +379,66 @@ go.app = function() {
             var registration_info = {
                 "reg_type": "momconnect_prebirth",
                 "registrant_id": self.im.user.answers.registrant.id,
+                "registration_type" : self.im.user.answers.registration_type,
```
rudigiesler: This isn't how we set the registration type on SEED. It's set on the `reg_type` field as either `momconnect_prebirth` or `whatsapp_prebirth`

go-app-ussd_chw.js
```diff
@@ -778,29 +821,65 @@ go.app = function() {
...
-                next: "state_save_subscription"
+                next: "state_set_registration_type"
+            });
+        });
+
+        self.add("state_set_registration_type", function(name){
+            /*var registration_types = {
+                'sms': 'SMS',
+                'whatsapp': 'WhatsApp',
+            };*/
```
rudigiesler: We can probably remove this commented out code

go-app-ussd_chw.js
OMosimege: Is there a way to differentiate a whatsapp subscription to a non-whatsapp one?

go-app-ussd_chw.js
OMosimege: Thanks for pointing this out

go-app-ussd_chw.js
rudigiesler: Yes, the `reg_type` field determines what kind of registration it is.

For the `momconnect_prebirth`, it's an SMS registration, for `whatsapp_prebirth`, is a whatsapp registration

OMosimege (commented):
👍

rudigiesler (approved):

PR #188, opened 2018-09-26
Default to WhatsApp registration for Clinic USSD line

rudigiesler:
This PR removes the channel selection screen, and will create a WhatsApp registration if the user is registered on WhatsApp, else it will create an SMS registration.

It also changes the Clinic USSD tests to use dynamic fixtures, so it might be useful to review the first commit separately to the rest of the commits.

### DIFF
--- a/test/ussd_clinic.test.js
+++ b/test/ussd_clinic.test.js
@@ -4,15 +4,19 @@ var assert = require('assert');
 var AppTester = vumigo.AppTester;
 
 var fixtures_IdentityStore = require('./fixtures_identity_store');
+var fixtures_IdentityStoreDynamic = require('./fixtures_identity_store_dynamic')();
 var fixtures_StageBasedMessaging = require('./fixtures_stage_based_messaging');
+var fixtures_StageBasedMessagingDynamic = require('./fixtures_stage_based_messaging_dynamic')();
 var fixtures_MessageSender = require('./fixtures_message_sender');
+var fixtures_MessageSenderDynamic = require('./fixtures_message_sender_dynamic')();
 var fixtures_Hub = require('./fixtures_hub');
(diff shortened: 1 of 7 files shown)


test/ussd_clinic.test.js
```diff
@@ -243,6 +308,22 @@ describe("app", function() {
...
+                                'msisdn': '+27820001001'
+                            }));
+                            api.http.fixtures.add(fixtures_StageBasedMessagingDynamic.active_subscriptions());
+                            api.http.fixtures.add(fixtures_JembiDynamic.momconnect_exists("123456", "test clinic"));
+                            api.http.fixtures.add(fixtures_MessageSenderDynamic.send_message({
+                                to_identity: "cb245673-aa41-4302-ac47-00000001002",
+                                content: "Please dial back in to *120*550*2# to complete the pregnancy registration."
+                            }));
+                            api.http.fixtures.add(fixtures_Pilot().exists({
+                                number: '+27123456789',
```
OMosimege, line 322: What number is this? `+27123456789`

rudigiesler: This is the number of our whatsapp account (the number that the user sees the message come from).

It's set in the application config at the top of the file in the test setup.

OMosimege: Is this the same for the CHW line?

OMosimege (approved):
Everything look great! 🍰

test/ussd_clinic.test.js
rudigiesler: Yes, this will be the same for all of the MomConnect USSD lines, since we use the same WhatsApp line for all MomConnect messages.

OMosimege: I see, thanks Rudi

PR #189, opened 2018-09-27
Default to WhatsApp registration for public USSD line

rudigiesler:
This PR removes the channel selection screen, defaulting to WhatsApp if the check returns true, otherwise defaulting to SMS.

It changes the content of the registration confirmation screen, and makes it channel specific.

It changes the content of the welcome SMS, and makes it channel specific.


OMosimege (approved):
Great job Rudi! 🍰

PR #190, opened 2018-09-27
Update translation files for Clinic USSD whatsapp default reg

rudigiesler:
The clinic line is english only, but the welcome SMS that it sends isn't.

This PR updates the clinic USSD translation files with the new content of the welcome SMS


OMosimege (approved):
Everything seems to be in order 👍

PR #193, opened 2018-10-02
Change PMTCT reg channel to be the same as existing registration

rudigiesler:
This PR changes the SEED PMTCT USSD line to create a PMTCT registration for the same channel as the user's current subscription is on.


OMosimege (approved):
Looks great Rudi :cake

PR #195, opened 2018-10-03
add user-agents to API calls

smn:
Adding this allows us to identify still using Wassup by looking at the logs and spotting the user-agents.


smn:
@rudigiesler & @erikh360 I no longer have the right node version installed to build this. Is the build done during a deploy ?

erikh360:
@smn we usually commit the `go-*` files, not sure if it will happen during the deploy so I've added them.

...and +1 wtih.

smn:
@erikh360 thanks!

PR #196, opened 2018-10-04
Feature/momza 1029 popi wa default reg

OMosimege:
This PR implements whatsapp default registration for the POPI user data line

### DIFF
--- a/src/ussd_popi_user_data.js
+++ b/src/ussd_popi_user_data.js
@@ -933,7 +933,7 @@ go.app = function() {
             return self.states.create("state_new_number_already_exists");
           }
           else{
-            return self.states.create("state_switch_number"); //used to be state_new_number_channel
+            return self.states.create("state_new_number_channel"); 
           }
         }
       );
@@ -961,10 +961,12 @@ go.app = function() {
         }).then(function(confirmed) {
             if(confirmed) {
                 channel = 'whatsapp';
+                self.map_channel(channel);
                 self.im.user.set_answer('channel', channel);
             }
             else{
                 channel = 'sms';
+                self.map_channel(channel);
                 self.im.user.set_answer('channel', channel);
             }
         }).then(function() {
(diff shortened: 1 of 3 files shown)


src/ussd_popi_user_data.js
```diff
@@ -317,7 +319,8 @@ go.app = function() {
                     new Choice('state_change_identity', $('Update my identification'))
                 ];
                 if(confirmed) {
-                    var alternate_channel = self.im.user.answers.channel == 'sms' ? 'whatsapp' : 'sms';
+                    //alternate_channel = 'whatsapp';
```
rudigiesler: We can probably remove this comment

src/ussd_popi_user_data.js
```diff
@@ -950,25 +953,44 @@ go.app = function() {
...
-                  new Choice('sms', $('SMS'))
-              ],
-              next: 'state_switch_number'
-          });
+          var msisdn = utils.normalize_msisdn(self.im.user.get_answer('state_enter_new_phone_number'), "27");
+          var channel;
+          return self.is_valid_recipient_for_pilot({ // msisdn whatsappable
+            address: msisdn,
+            wait: true,
+        }).then(function(confirmed) {
```
rudigiesler, line 960: The indentation from here on seems to be one indent too little.

src/ussd_popi_user_data.js
```diff
@@ -950,25 +953,44 @@ go.app = function() {
...
+            if(confirmed) {
+                channel = 'whatsapp';
+                self.map_channel(channel);
```
rudigiesler: `map_channel` just takes the channel type, and maps it to a user readable format. Since we're not doing anything with the result here, we can just remove the two `map_channel` calls.

src/ussd_popi_user_data.js
OMosimege: Thanks for noting this out

src/ussd_popi_user_data.js
OMosimege: ahh thanks

rudigiesler (approved):

PR #197, opened 2018-10-23
Move contact check to Engage API

rudigiesler:
Currently the contact checks are still on the old Wassup API, which is deprecated and scheduled to be turned off.

This PR moves the contact checks over to the new Engage API, and also extracts them into a separate module, removing a lot of duplicate code.


erikh360 (approved):
Nice!

PR #198, opened 2018-10-24
Fix HTTP headers for USSD app

rudigiesler:
In a previous PR, we added User Agent headers to http requests, so that we can track which apps are making requests. This was added incorrectly, the PR fixes that.

### DIFF
--- a/src/ussd_nurse.js
+++ b/src/ussd_nurse.js
@@ -26,7 +26,7 @@ go.app = function() {
         var engage;
 
         self.init = function() {
-            var config = {headers: {'User-Agent': 'Jsbox/NDoH-Nurse'}};
+            var config = {headers: [{'User-Agent': 'Jsbox/NDoH-Nurse'}]};
             // initialising services
             is = new SeedJsboxUtils.IdentityStore(
                 new JsonApi(self.im, config),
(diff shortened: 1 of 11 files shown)


src/ussd_nurse.js
```diff
@@ -26,7 +26,7 @@ go.app = function() {
         var engage;
 
         self.init = function() {
-            var config = {headers: {'User-Agent': 'Jsbox/NDoH-Nurse'}};
+            var config = {headers: [{'User-Agent': 'Jsbox/NDoH-Nurse'}]};
```
erikh360: Is the list in the right place here?

erikh360 (commented):
Just the one question.

src/ussd_nurse.js
rudigiesler: Ah, it's not, will fix

rudigiesler: Although we should probably remove this app, since we've moved over to RapidPro

rudigiesler:
@erikh360 Fixed. Ready for another review

erikh360 (approved):
Cool!, we should create a maintenance ticket to remove the NC things from seed.

PR #201, opened 2018-11-02
Feature/momza 1075 fix clinic ussd already registered number

OMosimege:
Fix for clinic USSD line, if the number entered already has an active momconnect subscription, then it shows an error message. But that error message shows the number that is dialling the line, not the number that was checked for an active subscription

Initially, it was using the operator msisdn to check the subscription, and not the registrant msisdn (which in most cases is different from the operator msisdn)

### DIFF
--- a/test/ussd_clinic.test.js
+++ b/test/ussd_clinic.test.js
@@ -574,6 +574,33 @@ describe("app", function() {
             });
         });
 
+        describe("when on state_mobile_no and number has no subscription", function() {
+            it("should go to state_consent", function() {
+                return tester
+                .setup(function(api) {
+                    api.http.fixtures.add(fixtures_IdentityStoreDynamic.identity_search({
+                        'msisdn': '+27820001001'
+                    }));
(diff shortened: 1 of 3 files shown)


test/ussd_clinic.test.js
```diff
@@ -574,6 +574,33 @@ describe("app", function() {
...
+                            'We need to collect, store & use her info. She ' +
+                            'may get messages on public holidays & weekends. ' +
+                            'Does she consent?',
+                            '1. Yes',
+                            '2. No'
+                    ].join('\n')
+                })
+                .run();
+            });
+        });
```
rudigiesler, line 602: I don't think that this test will trigger the bug. We need to test that the output of `state_already_subscribed` shows the entered number, not the current handset's number.

We can probably modify one of the existing tests to have a different `setup.user.addr` to what's configured in `registrant_msisdn`, and ensure that the response includes `registrant_msisdn`

OMosimege: Thanks will test for that as well

rudigiesler (approved):

PR #207, opened 2018-11-19
PMTCT pre to post missing baby dob fix

erikh360:
If a mom registers for prebirth and moves to postbirth, then tries to register for PMTCT it fails with missing baby date of birth.

In this case we should use the last edd we had.


OMosimege (approved):
Looks good! 🍰

PR #209, opened 2019-01-10
Fix tsonga translations for popi user data

rudigiesler:
For the Tsonga translations for the POPI User Data USSD line, the variable
names have also been translated, which means that it won't be able to render.


OMosimege (approved):
Looks good 👍

PR #210, opened 2019-01-10
Fix POPI User Data Afrikaans translation

rudigiesler:
One of the variables in the translation was also translated, leading to errors.


OMosimege (approved):
good work 👍

PR #211, opened 2019-01-17
Change to swt 2 for sms registrations

rudigiesler:
Currently we're sending an SWT of 3 for SMS registrations, which is actually
the Android software type. This PR changes this to the correct SWT for SMS,
    which is 2


OMosimege (approved):
Looks good 👍

PR #212, opened 2019-02-13
Feature/momza 1143 shorten auto responses

OMosimege:
Some of the autoresponses are too long and are being sent over two SMSes, this PR is to shorten those messages


messages.pot
```diff
@@ -0,0 +1,65 @@
+msgid ""
```
rudigiesler: We don't need the `messages.pot` file, that's just a temporary file to update the po files from. Once they're updated we don't have a need for it.

rudigiesler (approved):

PR #213, opened 2019-02-25
Feature/momza 1155 update message sender calls to use templates

OMosimege:
In this PR we:
Add templates for compliment and complaint instructions from Public USSD line
Add templates for PMTCT post-registration welcome messages

### DIFF
--- a/src/ussd_pmtct_seed.js
+++ b/src/ussd_pmtct_seed.js
@@ -542,11 +542,11 @@ go.app = function() {
         self.send_registration_thanks = function() {
             var lang_code = self.im.user.lang || "eng_ZA";
             var lang = engage.LANG_MAP[lang_code];
-            var content = self.im.user.i18n($(
+            var message1 = self.im.user.i18n($(
                 "HIV positive moms can have an HIV negative baby! You can get free " +
                 "medicine at the clinic to protect your baby and improve your health"
             ));
-            var text = self.im.user.i18n($(
+            var message2 = self.im.user.i18n($(
                 "Recently tested HIV positive? You are not alone, many other pregnant " +
                 "women go through this. Visit b-wise.mobi or call the AIDS Helpline " +
                 "0800 012 322"
@@ -558,14 +558,14 @@ go.app = function() {
                     return ms.create_outbound(
                         self.im.user.answers.identity.id,
                         self.im.user.answers.msisdn,
-                        content,
+                        message1,
                         {
                           channel: this.channel,
                           metadata : {
                             template: {
                               name: "important_info",
                               language: lang,
-                              variables: content
+                              variables: message1
                           }},
                         }
                     );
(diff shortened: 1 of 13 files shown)


src/ussd_pmtct_seed.js
```diff
@@ -564,6 +539,55 @@ go.app = function() {
...
         });
 
+        self.send_registration_thanks = function() {
+            var lang_code = self.im.user.lang || "eng_ZA";
+            var lang = engage.LANG_MAP[lang_code];
+            var content = self.im.user.i18n($(
+                "HIV positive moms can have an HIV negative baby! You can get free " +
+                "medicine at the clinic to protect your baby and improve your health"
+            ));
+            var text = self.im.user.i18n($(
```
rudigiesler: We can maybe rather name these something like `message1` and `message2`, so that it's a bit less confusing than `content` and `text`

src/ussd_public.js
```diff
@@ -312,37 +312,55 @@ go.app = function() {
         };
 
         self.send_compliment_instructions = function() {
+            var lang = engage.LANG_MAP[self.im.user.answers.state_language];
```
rudigiesler: In order to get to the compliment/complaint screens, the user doesn't go through the `state_language` screen, so this won't be set.

If you look at the `state_start`, if they're already registered, we use `self.im.user.set_lang` to set the language on `self.im.user.lang`, so that's a better place to find the user's language.

test/fixtures_message_sender.js
```diff
@@ -565,8 +565,13 @@ module.exports = function() {
                     "to_identity": "cb245673-aa41-4302-ac47-00000001002",
                     "to_addr": "+27820001002",
                     "content": "Please reply to this message with your compliment. If it relates to the service at the clinic, include the clinic or clinic worker name. Standard rates apply.",
-                    "metadata": {}
-                }
+                    "metadata":{
+                      "template":{
+                        "name":"important_info",
+                        "variables":"Please reply to this message with your compliment. If it relates to the service at the clinic, include the clinic or clinic worker name. Standard rates apply."
```
rudigiesler, line 572: I would also expect a `language` field to be inside the `template`, this is probably because of the issue of where we're getting the language from that I mentioned.

test/fixtures_message_sender.js
```diff
@@ -1017,32 +1085,82 @@ module.exports = function() {
...
+        // 147: post outbound message for cb245673-aa41-4302-ac47-00000000001
+          {
+              "key": "post.ms.outbound.27820001002",
+              "request": {
+                  "url": 'http://ms/api/v1/outbound/',
+                  "method": 'POST',
+                  "data": {
+                      "to_identity": "cb245673-aa41-4302-ac47-00000000002",
+                      "to_addr": "+27820000111",
+                      "content": "HIV positive moms can have an HIV negative baby! You can get free medicine at the clinic to protect your baby and improve your health",
```
rudigiesler: Since we're using the `api.http.fixtures.add` to add the fixture in the test, we don't need this fixture here (unless there's another test that's not using the dynamic fixtures)

rudigiesler (approved):

PR #216, opened 2019-04-02
Update language mapping

rudigiesler:
All languages except for English and Afrikaans need to be mapped to English
WhatsApp templates


OMosimege (commented):
👍

OMosimege (approved):

PR #218, opened 2019-04-25
Remove refer a friend from unregistered nurses.

rudigiesler:
We don't want unregistered users to be able to sign up other users. You have to be signed up first yourself to be able to refer a friend to the service


KaitCrawford (approved):
:+1: lgtm

PR #219, opened 2019-04-26
Block registering already registered friend

rudigiesler:
Currently, the refer-a-friend feature doesn't block a registration going through for an already-registered msisdn.

This PR blocks that registration from going through with an error message, asking the user to type in a different number.


KaitCrawford (approved):
:+1:

PR #220, opened 2019-04-30
Add facility code blacklist for NurseConnect

rudigiesler:
Adds a configurable blacklist for blocking certain facility codes from registration on NurseConnect


KaitCrawford (approved):
:+1:

PR #222, opened 2019-07-16
Remove welcome SMS from clinic registrations

rudigiesler:
https://github.com/praekeltfoundation/ndoh-hub/pull/239/ adds the welcome SMS to the Hub logic, so we no longer need to send it from the USSD application.


OMosimege (approved):
👍

PR #223, opened 2019-07-19
Send sid and eid to OpenHIM

rudigiesler:
We need to send the SystemID `sid` parameter (the Contact UUID) as well as the EventID `eid` parameter (any UUID unique to the event) in order for us to be able to retry requests and have deduplication.

In this case, since the request is done in the USSD session, we cannot afford to have retries, so we just generate a random UUID for the event ID.


OMosimege (approved):
This looks great 👍

PR #224, opened 2019-08-15
Use new API for validating and normalising MSISDNs

rudigiesler:
This will allow us to have better msisdn validation


OMosimege (approved):
👍

PR #226, opened 2019-10-08
Add basic setup for rapidpro clinic ussd

OMosimege:


src/ussd_clinic_rapidpro.js
```diff
@@ -0,0 +1,27 @@
...
+        var $ = self.$;
+
+        self.init = function() {
+        };
+
+        self.states.add("state_start", function(name) {
+            return new EndState(name, {
+                text: $(
+                    'Hello mom! You can reply to any MomConnect message with a question, compliment or complaint and ' +
+                    'our team of experts will get back to you.'
```
rudigiesler: I don't think we should be using the same text here as for the basic setup of the public line, otherwise we can't be sure that things are setup correctly. I think we  should rather use the text in the first screen of the clinic line here.

rudigiesler (approved):

PR #233, opened 2019-10-15
Create RapidPro Public USSD application

rudigiesler:
This implements the logic for the public USSD line that interacts with RapidPro


OMosimege (approved):
Great work 👍

PR #234, opened 2019-10-16
Send the source to the RapidPro flow

rudigiesler:
We want to be able to use this flow for any external public registration, so we need to state where this registration came from


OMosimege (approved):
👍

PR #235, opened 2019-11-08
Remove calls to get list of flows from RapidPro

rudigiesler:
Currently, we have enough flows that we're hitting the body size limits of HTTP requests in the sandbox.

These calls to get the list of flows isn't necessary, we can just put the flow ID into the application config, and skip the HTTP call


KaitCrawford (approved):
:+1: Looks good to me

PR #238, opened 2019-12-03
added timeout state, and tests

OMosimege:


go-app-ussd_public_rapidpro.js
```diff
@@ -182,6 +182,31 @@ go.app = function() {
...
+                    return creator(name, opts);
+
+                var timeout_opts = opts || {};
+                timeout_opts.name = name;
+                return self.states.create('state_timed_out', timeout_opts);
+            });
+        };
+
+        self.states.add('state_timed_out', function(name, creator_opts) {
+            return new ChoiceState(name, {
```
rudigiesler: We can probably use a `MenuState` here, instead of having a `next` function that returns `choice.value`

rudigiesler (commented):
Just a small comment, looks great otherwise!

rudigiesler (approved):

PR #244, opened 2019-12-19
Feature/mcmr 38 chw registration

OMosimege:
in this PR we create the CHW rapidpro USSD flow for registration

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -102,6 +102,9 @@ module.exports = function (grunt) {
                 ],
                 ussd_chw_rapidpro: [
                     'src/index.js',
+                    'src/engage.js',
+                    'src/rapidpro.js',
+                    'src/openhim.js',
                     '<%= paths.src.app.ussd_chw_rapidpro %>',
                     'src/init.js'
                 ],
@@ -214,6 +217,9 @@ module.exports = function (grunt) {
                 ],
                 ussd_chw_rapidpro: [
                     'test/setup.js',
+                    'src/engage.js',
+                    'src/rapidpro.js',
+                    'src/openhim.js',
                     '<%= paths.src.app.ussd_chw_rapidpro %>',
                     'test/ussd_chw_rapidpro.test.js'
                 ],
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -104,7 +104,6 @@ module.exports = function (grunt) {
                     'src/index.js',
                     'src/engage.js',
                     'src/rapidpro.js',
-                    'src/openhim.js',
                     '<%= paths.src.app.ussd_chw_rapidpro %>',
                     'src/init.js'
                 ],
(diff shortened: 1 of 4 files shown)


Gruntfile.js
```diff
@@ -102,6 +102,9 @@ module.exports = function (grunt) {
                 ],
                 ussd_chw_rapidpro: [
                     'src/index.js',
+                    'src/engage.js',
+                    'src/rapidpro.js',
+                    'src/openhim.js',
```
rudigiesler: It doesn't seem like we're using `openhim.js` in this application, so we can probably remove it here.

src/ussd_chw_rapidpro.js
```diff
@@ -1,30 +1,616 @@
...
+                            "the examples in the messages. Enter the mother's details."
+                        );
+                    }
+                },
+                next: "state_check_subscription"
+            });
+        });
+
+        self.add("state_check_subscription", function(name, opts) {
+            var msisdn = utils.normalize_msisdn(self.im.user.addr, "ZA");
```
rudigiesler: This might also be the address they entered in `state_enter_msisdn`, so if they went through that state and entered an msisdn, we should be using that instead of the msisdn that they're dialling from

src/ussd_chw_rapidpro.js
```diff
@@ -1,30 +1,616 @@
...
+            // Fire and forget a background whatsapp contact check
+            self.whatsapp.contact_check(msisdn, false).then(_.noop, _.noop);
+
+            return self.rapidpro.get_contact({urn: "tel:" + msisdn})
+                .then(function(contact) {
+                    self.im.user.set_answer("contact", contact);
+                    // Set the language if we have it
+                    if(_.isString(_.get(contact, "language"))) {
+                        return self.im.user.set_lang(contact.language);
```
rudigiesler: The CHW line is English only, and we don't want to set it to the mother's language either (since it's the CHW dialling the line, not the mother), so we can remove setting the language.

src/ussd_chw_rapidpro.js
```diff
@@ -1,30 +1,616 @@
...
+                    }
+                    return self.states.create(name, opts);
+                });
+        });
+
+        self.add("state_active_subscription", function(name) {
+            var msisdn = utils.readable_msisdn(
+                _.get(self.im.user.answers, "state_enter_msisdn", self.im.user.addr), "27");
+            var choices = [new Choice("state_enter_msisdn", $("Use a different number"))];
+            choices.push(new Choice("state_exit", $("Exit")));
```
rudigiesler: It's a little confusing that we're building the list of choices like this, when the list of choices will always be the same, so we can probably just add them as a list below, instead of building them here.

src/ussd_chw_rapidpro.js
```diff
@@ -1,30 +1,616 @@
...
+                    "Thank you for considering MomConnect. We respect her decision. Have a lovely day."
+                )
+            });
+        });
+
+        self.add("state_research_consent", function(name) {
+            // Skip this state if we already have consent
+            var consent = _.get(self.im.user.get_answer("contact"), "fields.research_consent");
+            if(consent === "TRUE") {
+                return self.states.create("state_opt_in");
```
rudigiesler: The next state here is `state_id_type`, not `state_opt_in`

rudigiesler (approved):

PR #246, opened 2019-12-23
added more information states along with tests

OMosimege:


go-app-ussd_chw_rapidpro.js
```diff
@@ -368,15 +369,15 @@ go.app = function() {
...
+                    "We need to process the mom's personal info to send her relevant messages about " +
                     "her pregnancy or baby. Does she agree?"
                 ),
                 error: $("Sorry, please reply with the number next to your answer. Does she agree?"),
                 accept_labels: true,
                 choices: [
                     new Choice("state_message_consent", $("Yes")),
                     new Choice("state_info_consent_denied", $("No")),
-                    new Choice("state_question_menu", $("I need more info to decide")),
+                    new Choice("state_question_menu", $("Needs more info to decide")),
```
rudigiesler, line 380: Does the original `She needs more info to decide` not fit in the character limit? If so, can you also make the same change to the flow diagram so that they match

rudigiesler (commented):
Just one small question, looks good otherwise

go-app-ussd_chw_rapidpro.js
OMosimege: yes, the full text takes it over the limit `163 > 160` I'll add a comment on the powerpoint

rudigiesler (approved):

PR #248, opened 2019-12-24
Add more information on rapidpro trigger for number change

rudigiesler:
This PR adds the `old_msisdn` and `source` fields, which we will need in the rapidpro flow.


OMosimege (approved):
looks good

PR #251, opened 2020-01-06
Feature/mcmr 41 create opt out

OMosimege:


src/ussd_optout_rapidpro.js
```diff
@@ -1,29 +1,333 @@
...
+                }).then(function() {
+                    // Delegate to the correct state depending on group membership
+                    var contact = self.im.user.get_answer("contact");
+                    if(self.contact_in_group(contact, self.im.config.public_group_ids)) {
+                        return self.states.create("state_opt_out");
+                    } else if(self.contact_in_group(contact, self.im.config.clinic_group_ids)){
+                        return self.states.create("state_opt_out");
+                    } else {
+                        return self.states.create("state_check_previous_optout");
+                    }
```
rudigiesler, line 72: I'm wondering if it might be better to have all of the logic in one place, so you can easily see where the user will end up, without having to check multiple states. So adding something like:
```
if(self.contact_in_group(contact, _.concat(self.im.config.public_group_ids, self.im.config.clinic_group_ids))){
    return self.states.create("state_opt_out");
} else if(self.contact_in_group(contact, self.im.config.optout_group_ids && _.get(contact, "fields.optout_reason") === "")) {
    return self.states.create("state_get_optout_reason");
} else {
    return self.states.create("state_no_previous_optout");
}
```
Where `state_get_optout_reason` is just `state_check_previous_optout` without the if/else logic.

src/ussd_optout_rapidpro.js
```diff
@@ -1,29 +1,333 @@
...
+                timeout_opts.name = name;
+                return self.states.create('state_timed_out', timeout_opts);
+            });
+        };
+
+        self.states.add('state_timed_out', function(name, creator_opts) {
+            return new MenuState(name, {
+                question: $('Welcome back. Please select an option:'),
+                choices: [
+                    new Choice(creator_opts.name, $('Continue signing up for messages')),
```
rudigiesler: This doesn't really make sense, they're not signing up for messages on this line.

src/ussd_optout_rapidpro.js
```diff
@@ -1,29 +1,333 @@
...
                     new Choice("miscarriage", $("Miscarriage")),
-                    new Choice("stillbirth", $("Baby was stillborn")),
-                    new Choice("babyloss", $("Baby died")),
-                    new Choice("not_useful", $("Messages not useful")),
-                    new Choice("other", $("Other"))
+                    new Choice("stillborn", $("Baby was stillborn")),
+                    new Choice("babyloss", $("Baby passed away")),
+                    new Choice("not helpful", $("Msgs aren't helpful")),
+                    new Choice("other", $("Other")),
+                    new Choice("not stated", $("I prefer not to say"))
```
rudigiesler: I think we should stick with the existing reasons that we're already using in the event store, so that we don't have to translate between the two when storing this, and when submitting to DHIS2: https://github.com/praekeltfoundation/ndoh-hub/blob/develop/eventstore/models.py#L35-L41

src/ussd_optout_rapidpro.js
```diff
@@ -1,29 +1,333 @@
...
+                ),
+                choices: [
+                    new Choice("yes", $("Yes")),
+                    new Choice("no", $("No"))
+                ],
+                next: function(choice) {
+                    if(choice.value === "yes") {
+                        return "state_info_deleted";
+                    } else {
+                        return self.states.create("state_trigger_rapidpro_flow");
```
rudigiesler: We can just return the string `"state_trigger_rapidpro_flow"` here, like we do for `state_info_deleted`

src/ussd_optout_rapidpro.js
```diff
@@ -1,29 +1,333 @@
...
+                    accept_labels: true,
+                    options_per_page: null,
+                    characters_per_page: 160,
+                    choices: [
+                        new Choice("miscarriage", $("Miscarriage")),
+                        new Choice("stillborn", $("Baby was stillborn")),
+                        new Choice("babyloss", $("Baby passed away")),
+                        new Choice("no", $("Msgs aren't helpful")),
+                        new Choice("no", $("Other")),
+                        new Choice("no", $("I prefer not to say"))
```
rudigiesler: We probably want these 3 to still have a value other than `no`, so that we know what the reason is, and we can send this reason to RapidPro when we trigger the flow.

src/ussd_optout_rapidpro.js
```diff
@@ -1,29 +1,333 @@
...
+            var msisdn = utils.normalize_msisdn(self.im.user.addr, "ZA");
+            return self.rapidpro.start_flow(self.im.config.flow_uuid, null, "tel:" + msisdn, {
+                babyloss_subscription: 
+                    self.im.user.get_answer("state_loss_optout") === "yes" ? "TRUE" : "FALSE",
+                delete_info_for_babyloss: 
+                    self.im.user.get_answer("state_delete_info_for_loss") === "yes" ? "TRUE" : "FALSE",
+                delete_info_consent: 
+                        self.im.user.get_answer("state_delete_research_info") === "yes" ? "TRUE" : "FALSE",
+                optout_reason: self.im.user.get_answer("optout_reason"),
+                language: self.im.user.lang,
```
rudigiesler: The user isn't changing there language in this flow, so we don't need to send the user's language to RapidPro, RapidPro already knows what the user's language is

src/ussd_optout_rapidpro.js
```diff
@@ -1,29 +1,333 @@
...
+                }else if (self.im.user.get_answer("state_check_previous_optout") === "no"){
+                    return self.states.create("state_nonloss_optout");
+                }else if (self.im.user.get_answer("state_delete_info_for_loss") === "no"){
+                    return self.states.create("state_loss_subscription_with_info_retainment");
+                }else if (self.im.user.get_answer("state_info_deleted") === "yes" ){
+                    return self.states.create("state_nonloss_optout");
+                }else if (self.im.user.get_answer("state_delete_research_info") === "no" ){
+                    return self.states.create("state_nonloss_optout");
+                } else{
+                    return self.states.create("state_nonloss_optout");
```
rudigiesler, line 255: It seems like `state_nonloss_optout` is repeated quite a bit, so we can probably just do the checks for `state_loss_subscription_without_info_retainment` and `state_loss_subscription_with_info_retainment`, and if it's neither of those two, just do an
```
} else {
    return self.states.create("state_nonloss_optout"); 
}
```

test/ussd_optout_rapidpro.test.js
```diff
@@ -8,25 +10,748 @@ describe("ussd_optout_rapidpro app", function() {
...
+                .input("2")
+                .check.interaction({
+                    state: "state_no_optout",
+                    reply: 
+                        "Thanks! MomConnect will continue to send you helpful messages and process your " +
+                        "personal info. Have a lovely day!"
+                })
+                .run();
+        });
+        it("should go to state_delete_research_info if they choose yes", function() {
```
rudigiesler: should go to state_optout_reason if they choose yes

test/ussd_optout_rapidpro.test.js
```diff
@@ -8,25 +10,748 @@ describe("ussd_optout_rapidpro app", function() {
...
+                .run();
+        });
+        it("should display an error if an incorrect input is sent", function() {
+            return tester
+                .setup.user.state("state_optout_reason")
+                .input("foo")
+                .check.interaction({
+                    state: "state_optout_reason",
+                    reply: [
+                        "We'll stop sending msgs. Why do you want to stop your MC msgs?",
```
rudigiesler, line 293: This error doesn't seem to be any different to the question. If this is intentional, maybe we should change the test description to `should display the question if an incorrect input is sent` to make that clear.

src/ussd_optout_rapidpro.js
OMosimege: thanks for picking this up, forgot to change the line

src/ussd_optout_rapidpro.js
OMosimege: thanks. Was not sure what to use

OMosimege: So I could change the choices to this? 
```
new Choice("Miscarriage", $("Miscarriage")),
                    new Choice("Stillborn", $("Baby was stillborn")),
                    new Choice("Babyloss", $("Baby passed away")),
                    new Choice("Not useful", $("Msgs aren't helpful")),
                    new Choice("Other", $("Other")),
                    new Choice("Unknown", $("I prefer not to say"))
```

src/ussd_optout_rapidpro.js
OMosimege: okay, thanks

src/ussd_optout_rapidpro.js
OMosimege: okay, will do

src/ussd_optout_rapidpro.js
OMosimege: will remove it

test/ussd_optout_rapidpro.test.js
OMosimege: ahh, thanks

test/ussd_optout_rapidpro.test.js
OMosimege: yes, that was the intention

src/ussd_optout_rapidpro.js
rudigiesler: I think we need to stick to the machine versions, `not_useful`, `babyloss`, etc, rather than the human-readable versions.

src/ussd_optout_rapidpro.js
OMosimege: will use this suggestion, thank you

src/ussd_optout_rapidpro.js
OMosimege: thanks for picking this up

rudigiesler (approved):

src/ussd_optout_rapidpro.js
```diff
@@ -1,29 +1,317 @@
...
+                accept_labels: true,
+                options_per_page: null,
+                characters_per_page: 160,
+                choices: [
+                    new Choice("Miscarriage", $("Miscarriage")),
+                    new Choice("Stillborn", $("Baby was stillborn")),
+                    new Choice("Babyloss", $("Baby passed away")),
+                    new Choice("Not useful", $("Msgs aren't helpful")),
+                    new Choice("Other", $("Other")),
+                    new Choice("Unknown", $("I prefer not to say"))
```
rudigiesler: We need to be using the machine versions, not the human readable versions, for the labels, e.g. `not_useful` instead of `Not useful`

src/ussd_optout_rapidpro.js
```diff
@@ -1,29 +1,317 @@
...
+                    accept_labels: true,
+                    options_per_page: null,
+                    characters_per_page: 160,
+                    choices: [
+                        new Choice("Miscarriage", $("Miscarriage")),
+                        new Choice("Stillborn", $("Baby was stillborn")),
+                        new Choice("Babyloss", $("Baby passed away")),
+                        new Choice("Not useful", $("Msgs aren't helpful")),
+                        new Choice("Other", $("Other")),
+                        new Choice("Unknown", $("I prefer not to say"))
```
rudigiesler: Same here for the labels

rudigiesler (approved):

PR #252, opened 2020-01-08
Add change research consent states

rudigiesler:
This will allow the user to change their research consent on the POPI USSD line


OMosimege (approved):
👍

PR #253, opened 2020-01-10
POPI add states for optout

rudigiesler:


src/ussd_popi_rapidpro.js
```diff
@@ -793,6 +793,193 @@ go.app = function() {
...
+
+        self.add("state_opt_out_reason", function(name) {
+            return new PaginatedChoiceState(name, {
+                question: $("We'll stop sending msgs. Why do you want to stop your MC msgs?"),
+                error: $("Sorry we don't understand. Please try again."),
+                choices: [
+                    new Choice("miscarriage", $("Miscarriage")),
+                    new Choice("stillbirth", $("Baby was stillborn")),
+                    new Choice("babyloss", $("Baby passed away")),
+                    new Choice("not_useful", $("Msgs aren't helpful")),
```
OMosimege, line 835: Our reasons match, except in the optout line the `not_useful` is a `not useful`, and everything starts with a capital letter, should I change that on the optout line?

OMosimege (approved):
Approved 👍 Just added one question regarding the optout line

src/ussd_popi_rapidpro.js
rudigiesler: Yes, I think I did put a comment on your PR about that. We want our reasons to match https://github.com/praekeltfoundation/ndoh-hub/blob/develop/eventstore/models.py#L35-L42 , so that we can just send it through without having to convert it first

PR #254, opened 2020-01-14
Fix null contact field handling

rudigiesler:
If RapidPro doesn't have a value for a field for a contact, then it will still have that field in the API response, but its value will be `null`, so we should handle that.


OMosimege (approved):
👍

PR #255, opened 2020-01-17
Add states for no sim msisdn change

rudigiesler:
This will allow a user to change their phone number if they no longer have access to their old SIM card


OMosimege (approved):
:+1

PR #257, opened 2020-01-22
Feature/mcmr 43 create pmtct registration

OMosimege:
This PR creates PMTCT registration, PMTCT loss switches, and PMTCT optout

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -116,7 +116,6 @@ module.exports = function (grunt) {
                 ],
                 ussd_pmtct_rapidpro: [
                     'src/index.js',
-                    'src/engage.js',
                     'src/rapidpro.js',
                     '<%= paths.src.app.ussd_pmtct_rapidpro %>',
                     'src/init.js'
@@ -236,7 +235,6 @@ module.exports = function (grunt) {
                 ussd_pmtct_rapidpro: [
                     'test/setup.js',
                     'src/rapidpro.js',
-                    'src/engage.js',
                     '<%= paths.src.app.ussd_pmtct_rapidpro %>',
                     'test/ussd_pmtct_rapidpro.test.js'
                 ],
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -116,6 +116,8 @@ module.exports = function (grunt) {
                 ],
                 ussd_pmtct_rapidpro: [
                     'src/index.js',
+                    'src/engage.js',
+                    'src/rapidpro.js',
                     '<%= paths.src.app.ussd_pmtct_rapidpro %>',
                     'src/init.js'
                 ],
@@ -226,12 +228,15 @@ module.exports = function (grunt) {
                 ],
                 ussd_optout_rapidpro: [
                     'test/setup.js',
+                    'src/engage.js',
                     'src/rapidpro.js',
                     '<%= paths.src.app.ussd_optout_rapidpro %>',
                     'test/ussd_optout_rapidpro.test.js'
                 ],
                 ussd_pmtct_rapidpro: [
                     'test/setup.js',
+                    'src/rapidpro.js',
+                    'src/engage.js',
                     '<%= paths.src.app.ussd_pmtct_rapidpro %>',
                     'test/ussd_pmtct_rapidpro.test.js'
                 ],
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -227,7 +227,6 @@ module.exports = function (grunt) {
                 ],
                 ussd_optout_rapidpro: [
                     'test/setup.js',
-                    'src/engage.js',
                     'src/rapidpro.js',
                     '<%= paths.src.app.ussd_optout_rapidpro %>',
                     'test/ussd_optout_rapidpro.test.js'
--- a/src/ussd_pmtct_rapidpro.js
+++ b/src/ussd_pmtct_rapidpro.js
@@ -105,7 +105,7 @@ go.app = function() {
                 }).then(function() {
                     // Delegate to the correct state depending on group membership
                     var contact = self.im.user.get_answer("contact");
-                    if(self.contact_in_group(contact, _.concat(self.im.config.public_group_ids, self.im.config.clinic_group_ids))){
+                    if(self.contact_in_group(contact, self.im.config.clinic_group_ids)){
                         if(self.contact_in_group(contact, self.im.config.pmtct_group_ids)){
                             return self.states.create("state_optout");
                         } else {
@@ -126,16 +126,23 @@ go.app = function() {
         });
 
         self.states.add("state_optout", function(name) {
-            return new MenuState(name, {
+            return new ChoiceState(name, {
                 question: "The mother is currently receiving messages about keeping her baby " +
                             "HIV-negative. Does she want to stop getting these messages?",
                 error:
                     "Sorry, please reply with the number next to your answer. " +
                     "Does she want to stop getting these messages?",
                 choices: [
-                    new Choice("state_optout_reason", "Yes"),
-                    new Choice("state_no_optout", "No")
+                    new Choice("yes", "Yes"),
+                    new Choice("no", "No")
                 ],
+                next: function(choice) {
+                    if(choice.value === "yes") {
+                        return "state_optout_reason";
+                    } else {
+                        return "state_no_optout";
+                    }
+                }
             });
         });
 
@@ -333,8 +340,9 @@ go.app = function() {
             return new EndState(name, {
                 next: "state_start",
                 text: $(
-                    "Thank you. The mother will receive messages about " + 
-                    "keeping her baby HIV-negative. Have a lovely day."
+                    "Welcome to the Department of Health’s MomConnect. To get msgs " +
+                    "about keeping your baby HIV-negative, register to MomConnect by " +
+                    "dialing *154*550*2# at the clinic."
                 )
             });
         });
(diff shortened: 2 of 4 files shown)


Gruntfile.js
```diff
@@ -226,12 +227,14 @@ module.exports = function (grunt) {
                 ],
                 ussd_optout_rapidpro: [
                     'test/setup.js',
+                    'src/engage.js',
```
rudigiesler: I don't think that we need `engage.js` on the rapidpro optout ussd line

src/ussd_pmtct_rapidpro.js
```diff
@@ -1,24 +1,393 @@
...
+                choices: [
+                    new Choice("yes", "Yes"),
+                    new Choice("no", "No")
+                ],
+                next: function(choice) {
+                    if(choice.value === "yes") {
+                        return "state_loss_subscription";
+                    } else {
+                        return "state_opted_out";
+                    }
```
rudigiesler: There's no real logic in here, so we can probably use a `MenuState` here instead, and simplify it a bit.

src/ussd_pmtct_rapidpro.js
```diff
@@ -1,24 +1,393 @@
...
+            });
+        });
+
+        self.states.add("state_opted_out", function(name) {
+            return new EndState(name, {
+                next: "state_start",
+                text:
+                    "Thank you. She will no longer receive messages from us about HIV. " +
+                    "For any medical concerns, please visit a clinic."
+            });
```
rudigiesler, line 191: We need to send this optout to RapidPro so that we can actually do the logic so that they will stop getting messages.

src/ussd_pmtct_rapidpro.js
```diff
@@ -1,24 +1,393 @@
...
+                    }
+                }
+            });
+        });
+
+        self.states.add("state_loss_subscription", function(name) {
+            return new EndState(name, {
+                next: "state_start",
+                text:
+                    "Thank you. She will receive messages of support from MomConnect in the coming weeks."
```
rudigiesler: We need to send this loss switch to RapidPro, so that we can do the logic to opt out the user, and subscribe them to loss messages.

src/ussd_pmtct_rapidpro.js
```diff
@@ -1,24 +1,393 @@
...
+                text: "Thank you for using MomConnect. Have a lovely day."
+            });
+        });
+
+        self.states.add("state_no_subscription", function(name) {
+            return new EndState(name, {
+                next: "state_start",
+                text: $(
+                    "Thank you. The mother will receive messages about " + 
+                    "keeping her baby HIV-negative. Have a lovely day."
```
rudigiesler: Should this state not be the following content? (It's if they're not registered on MomConnect):
```
Welcome to the Department of Health’s MomConnect. To get msgs about keeping your baby HIV-negative, register to MomConnect by dialing *154*550*2# at the clinic.
```

src/ussd_pmtct_rapidpro.js
```diff
@@ -1,24 +1,393 @@
...
+                .then(function(contact) {
+                    self.im.user.set_answer("contact", contact);
+                    // Set the language if we have it
+                    if(_.isString(_.get(contact, "language"))) {
+                        return self.im.user.set_lang(contact.language);
+                    }
+                }).then(function() {
+                    // Delegate to the correct state depending on group membership
+                    var contact = self.im.user.get_answer("contact");
+                    if(self.contact_in_group(contact, _.concat(self.im.config.public_group_ids, self.im.config.clinic_group_ids))){
```
rudigiesler: I think we only want clinic registrations to be able to register for PMTCT, there are no PMTCT messages for public registrations.

src/ussd_pmtct_rapidpro.js
OMosimege: so we save it as part of the data we send to rapidpro?

src/ussd_pmtct_rapidpro.js
OMosimege: thanks, I missed this error

src/ussd_pmtct_rapidpro.js
OMosimege: thanks, I missed this error

src/ussd_pmtct_rapidpro.js
OMosimege: okay did not know this, will remove the public subscriptions

src/ussd_pmtct_rapidpro.js
OMosimege: looking back, I wanted this to be a choice state so I can save the loss subscription result

Gruntfile.js
OMosimege: removed

src/ussd_pmtct_rapidpro.js
rudigiesler: Well if the user goes through this part of the flow, at the moment they won't ever reach the "send to RapidPro" state. The existing submission only happens on new registrations, not on opt outs or loss switches. So we either need to send them to the same flow, and have a variable that tells the flow whether this is a registration, optout, or loss switch, or we send the user to a different flow for each action.

src/ussd_pmtct_rapidpro.js
```diff
@@ -1,24 +1,413 @@
...
+                    new Choice("not_hiv_positive", "She's not HIV+"),
+                    new Choice("miscarriage", "Miscarriage"),
+                    new Choice("stillborn", "Baby was stillborn"),
+                    new Choice("babyloss", "Baby passed away"),
+                    new Choice("not_useful", "Msgs aren't helpful"),
+                    new Choice("other", "Other"),
+                    new Choice("unknown", "I prefer not to say")
+                ],
+                next: function(choice) {
+                    self.im.user.set_answer("optout_reason", choice.value);
```
rudigiesler: This will automatically be saved under `state_optout_reason`, so we don't need to save it here again.

src/ussd_pmtct_rapidpro.js
```diff
@@ -1,24 +1,413 @@
...
+                    if(_.includes(["miscarriage", "stillborn", "babyloss"], choice.value)) {
+                        return "state_loss_optout";
+                    } else {
+                        return "state_opted_out";
```
rudigiesler: We need to be submitting the optout to RapidPro here, not going to the end state

src/ussd_pmtct_rapidpro.js
```diff
@@ -1,24 +1,413 @@
...
+                    "Would she like to receive a small set of MomConnect messages?",
+                choices: [
+                    new Choice("yes", "Yes"),
+                    new Choice("no", "No")
+                ],
+                next: function(choice) {
+                    if(choice.value === "yes") {
+                        return "state_loss_subscription";
+                    } else {
+                        return "state_opted_out";
```
rudigiesler: We need to be submitting the optout to RapidPro here, not going to the end state

src/ussd_pmtct_rapidpro.js
```diff
@@ -1,24 +1,413 @@
...
+                error:
+                    "Sorry, please reply with the number next to your answer. " +
+                    "Would she like to receive a small set of MomConnect messages?",
+                choices: [
+                    new Choice("yes", "Yes"),
+                    new Choice("no", "No")
+                ],
+                next: function(choice) {
+                    if(choice.value === "yes") {
+                        return "state_loss_subscription";
```
rudigiesler: We need to be submitting the loss switch to RapidPro here, not going to the end state

src/ussd_pmtct_rapidpro.js
```diff
@@ -1,24 +1,413 @@
...
+                        return self.states.create("__error__");
+                    }
+                    return self.states.create(name, opts);
+                });
+        });
+
+        self.states.add("state_optout", function(name) {
+            return new ChoiceState(name, {
+                question: "The mother is currently receiving messages about keeping her baby " +
+                            "HIV-negative. Does she want to stop getting these messages?",
```
rudigiesler: I've noticed that a few of the user strings aren't surrounded by `$()`, which means that we won't be able to translate them. Any of the strings after the `self.im.user.set_lang(contact.language)` needs to be surrounded in `$()`, otherwise they won't be displayed in the user's language.

src/ussd_pmtct_rapidpro.js
OMosimege: thanks

src/ussd_pmtct_rapidpro.js
OMosimege: ahh yes, thanks for picking this up, forgot I changed a few things around

src/ussd_pmtct_rapidpro.js
OMosimege: same as above

src/ussd_pmtct_rapidpro.js
OMosimege: but according to the flow, that only happens after the `no active PMTCT` part of the flow?

src/ussd_pmtct_rapidpro.js
OMosimege: same as above, thanks

src/ussd_pmtct_rapidpro.js
rudigiesler: I think that should be moved up, because we do know their language for the opt out. In the code we're setting their language in `state_check_subscription`, so any state that comes after that must use the translation strings, and any state before that mustn't.

So either we can change where we set the language, or we update the flow to set the language where we're setting it in the code.

rudigiesler (approved):

PR #258, opened 2020-01-31
Translations for RapidPro

rudigiesler:
This PR is so that we can collect all the strings that require translation


OMosimege (approved):
👍

PR #260, opened 2020-02-11
Use WhatsApp URN type

rudigiesler:
We need to use the WhatsApp URN type for API calls to RapidPro, instead of the tel URN typ


OMosimege (approved):
👍

PR #266, opened 2020-03-05
added changes to PMTCT ussd screens

OMosimege:


go-app-ussd_pmtct_rapidpro.js
```diff
@@ -169,14 +169,14 @@ go.app = function() {
         self.add("state_enter_msisdn", function(name) {
             return new FreeText(name, {
                 question:
-                    "Please enter the cell number of the mother who would like to sign up to " +
-                    "receive HIV-related messages from MomConnect e.g. 0813547654.",
+                    "Please enter the cell number of the mom that would like to sign up to or " +
```
erikh360, line 172: cell number changed to no. earlier, should that be happening here too?

go-app-ussd_pmtct_rapidpro.js
```diff
@@ -169,14 +169,14 @@ go.app = function() {
...
+                    "opt out of HIV-related messages from MomConnect e.g. 0813547654.",
                 check: function(content) {
                     if(!utils.is_valid_msisdn(content, "ZA")) {
                         return (
-                            "Sorry, we don't understand that cell number. Please enter 10 digit " +
-                            "cell number that the mother would like to get HIV-related messages " +
-                            "on, e.g. 0813547654.");
+                            "Sorry, we don't understand that number. Please enter 10 digit " +
```
erikh360, line 177: hmm over here we've got number, instead of no. or cell number.

erikh360 (commented):
Looks good, technically I'm +1

Just a few question on the actual copy

go-app-ussd_pmtct_rapidpro.js
OMosimege: I did a copy and paste of how SxD did the screens, I could double check with them if this format was deliberate

go-app-ussd_pmtct_rapidpro.js
OMosimege: with this, there was a character limit issue, But we further state that we need the cell number in the next sentence `Please enter 10 digit cell number that the mother would like to get or opt out of HIV-related messages ` it clarifies to the user what is needed as input

go-app-ussd_pmtct_rapidpro.js
OMosimege: The change of the use of number to `no.` is used when there are character limitations, only 160 char can be used per screen. So its a decision they took as a standard across all of the USSD lines & message sets

erikh360 (approved):

PR #270, opened 2020-04-01
Initial CCMDD WC address update app

erikh360:
Start of the CCMDD WC address update app. Not complete I just want to deploy something to test the USSD lines.

The USSD flow is still being designed.


OMosimege (approved):
Setup looks good 👍

PR #273, opened 2020-04-02
Rest of the address update ussd app

erikh360:
The test coverage isn't great but it is just a prototype. We'll add some more when time allows

### DIFF
--- a/go-app-ussd_ccmdd_wc_address_update.js
+++ b/go-app-ussd_ccmdd_wc_address_update.js
@@ -321,8 +321,7 @@ go.app = (function() {
             !dob.isBetween(
               today.clone().add(-130, "years"),
               today.clone().add(-5, "years")
-            ) ||
-            _.parseInt(match[2]) >= 5000
+            )
           ) {
             return $(
               "Sorry, we don’t understand. Please try again by entering the your 13 digit South African ID number."
@@ -609,7 +608,7 @@ go.app = (function() {
       return new FreeText(name, {
         question: $("Please reply with your house number, e.g. 17."),
         check: function(content) {
-          var match = content.match(/^(\d{1,7})$/);
+          var match = content.match(/^(\d{1,7}[a-zA-Z]{0,1})$/);
           if (!match) {
             return $(
               "Sorry, we don’t understand. Please reply with your house " +
@@ -708,7 +707,7 @@ go.app = (function() {
         next: "state_start",
         text: $(
           "Thank you. You will receive your medication at the end of the " +
-            "month. The driver will contact you before the time.."
+            "month. The driver will contact you before the time."
         ).context({ msisdn: msisdn })
       });
     });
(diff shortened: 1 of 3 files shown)


go-app-ussd_ccmdd_wc_address_update.js
```diff
@@ -253,6 +273,354 @@ go.app = (function() {
...
+          if (
+            !match ||
+            !validLuhn(content) ||
+            !(dob = new moment(match[1], "YYMMDD")) ||
+            !dob.isValid() ||
+            !dob.isBetween(
+              today.clone().add(-130, "years"),
+              today.clone().add(-5, "years")
+            ) ||
+            _.parseInt(match[2]) >= 5000
```
rudigiesler: This will restrict the service to only female ID numbers. I'm not sure that we want to do that here.

go-app-ussd_ccmdd_wc_address_update.js
```diff
@@ -253,6 +273,354 @@ go.app = (function() {
...
+        },
+        next: "state_street_number"
+      });
+    });
+
+    self.add("state_street_number", function(name) {
+      return new FreeText(name, {
+        question: $("Please reply with your house number, e.g. 17."),
+        check: function(content) {
+          var match = content.match(/^(\d{1,7})$/);
```
rudigiesler: I can see some issues with this being too strict, eg. my number is `32B`

go-app-ussd_ccmdd_wc_address_update.js
```diff
@@ -312,8 +707,8 @@ go.app = (function() {
       return new EndState(name, {
         next: "state_start",
         text: $(
-          "Thank you. You will receive your medication at the end of the month. " +
-            "The driver will contact you before the time."
+          "Thank you. You will receive your medication at the end of the " +
+            "month. The driver will contact you before the time.."
```
rudigiesler: Extra `.` at the end

test/ussd_ccmdd_wc_address_update.test.js
```diff
@@ -1,61 +1,473 @@
...
+        .input({ session_event: "continue" })
+        .check.interaction({
+          state: "state_info_consent",
+          reply: [
+            "To have your prescription medication delivered to your door, we need to process your personal info.",
+            "Do you want to continue?",
+            "1. Yes",
+            "2. No"
+          ].join("\n"),
+          char_limit: 180
```
rudigiesler: Not `160`?

OMosimege (approved):
I see there is no test for state_timed_out, otherwise, looks great 👍

go-app-ussd_ccmdd_wc_address_update.js
erikh360: hehe, I live at 2A and thought about this

PR #278, opened 2020-04-02
Add name and surname and cleanup tests

erikh360:


go-app-ussd_ccmdd_wc_address_update.js
```diff
@@ -273,6 +273,36 @@ go.app = (function() {
       });
     });
 
+    self.add("state_first_name", function(name) {
+      return new FreeText(name, {
+        question: $("What is your first name?"),
+        check: function(content) {
+          if (!content.match(/^\w+$/)) {
```
rudigiesler: Can we check if this will match certain characters like `ê`?

rudigiesler (commented):
Just one tiny thing, probably not important for this stage, but just wanted to make sure it was thought of before we go live

go-app-ussd_ccmdd_wc_address_update.js
erikh360: Ah thanks, fixed that and removed another state, ready for another look!

rudigiesler (approved):

PR #279, opened 2020-04-02
Alternate ending for no tracing

rudigiesler:
If the user chooses not to be contacted, then we want different content to be displayed, as well as a continuation


OMosimege (approved):
:+1

PR #287, opened 2020-04-06
Remove WA and replace rapidpro with eventstore

erikh360:
Sorry about all the formatting changes


src/ussd_ccmdd_wc_address_update.js
```diff
@@ -457,88 +416,58 @@ go.app = (function() {
...
-        .start_flow(
-          self.im.config.flow_uuid,
-          null,
-          "whatsapp:" + _.trim(msisdn, "+"),
-          data
-        )
-        .then(function() {
-          return self.states.create("state_update_complete");
+    self.add("state_submit_data", function (name, opts) {
+      return new JsonApi(self.im)
```
rudigiesler, line 424: I think we need to also add authorisation headers to this request.

PR #291, opened 2020-04-07
CDU address update validation and msisdn added

erikh360:


go-app-ussd_ccmdd_wc_address_update.js
```diff
@@ -573,6 +590,7 @@ go.app = (function () {
       return new JsonApi(self.im)
         .post(self.im.config.eventstore.url + "/api/v2/cduaddressupdate/", {
           data: {
+            msisdn: self.im.user.addr,
```
rudigiesler: I've noticed that this is in the format `27820001001` instead of the international format `+27820001001`. I'm going to add something server side to format it correctly, but just want to make sure you're aware of that.

There is a helper function for formatting the address in international format, if you want to do it client side: https://github.com/praekeltfoundation/ndoh-jsbox/blob/develop/src/ussd_clinic_rapidpro.js#L129-L130

rudigiesler (approved):

PR #299, opened 2020-04-21
combined message consent + info consent states together

OMosimege:
In This PR, copy changes are done to the clinic ussd line where the message consent state is removed, and one consent is requested in the info consent state


rudigiesler (approved):
👍

PR #300, opened 2020-04-21
complete sassa ussd registration

erikh360:


src/ussd_sassa_registration.js
```diff
@@ -67,109 +93,389 @@ go.app = (function () {
...
+                      return $(
+                        [
+                          "An application for {{ id_number }} has already been made. The status is: {{ status }}.",
+                          "To appeal, call 0800 002 9999",
+                        ].join("\n")
+                      ).context({
+                        id_number: content,
+                        status: response.data.status,
+                      });
+                    }
```
rudigiesler, line 189: We should probably also have some error handling for if the request fails. Maybe not right now, but at some point.

src/ussd_sassa_registration.js
```diff
@@ -67,109 +93,389 @@ go.app = (function () {
...
+    });
+
+    self.add("state_name", function (name) {
+      return new FreeText(name, {
+        question: $(
+          "Please enter the applicant's name and surname as per identity " +
+            "document (eg John Doe)"
+        ),
+        check: function (content) {
+          if (content.length < 2 || content.indexOf(" ") == -1) {
```
rudigiesler: Maybe we should do a `.trim()` before checking length, so that two spaces doesn't match? Or maybe a regex of `\S+\s\S+.*`

rudigiesler: We should probably also do a length limitation or truncation, so that we don't hit our db field limit, or the character limit for the confirm screen

rudigiesler (approved):

PR #310, opened 2020-05-19
Add rapidpro flow trigger for SMS

rudigiesler:
This will trigger a RapidPro flow with the correct information required to send the receipt SMS. RapidPro will then handle creating the SMS content, and sending the SMS to the user

### DIFF
--- a/go-app-ussd_higherhealth_healthcheck.js
+++ b/go-app-ussd_higherhealth_healthcheck.js
@@ -1,9 +1,111 @@
 var go = {};
 go;
 
+go.RapidPro = function() {
+    var vumigo = require('vumigo_v02');
+    var url_utils = require('url');
+    var events = vumigo.events;
+    var Eventable = events.Eventable;
+
+    var RapidPro = Eventable.extend(function(self, json_api, base_url, auth_token) {
+        self.json_api = json_api;
+        self.base_url = base_url;
+        self.auth_token = auth_token;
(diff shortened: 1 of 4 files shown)


moh-moola (approved):
Looks solid to me 👍

go-app-ussd_higherhealth_healthcheck.js
```diff
@@ -367,6 +469,37 @@ go.app = (function () {
           "User-Agent": ["Jsbox/Covid19-Triage-USSD"]
         }
       }).then(function () {
+        return self.states.create("state_submit_sms");
+      }, function (e) {
+        // Go to error state after 3 failed HTTP requests
+        opts.http_error_count = _.get(opts, "http_error_count", 0) + 1;
```
miltontony, line 475: @rudigiesler I'm assuming that even though the USSD session itself might time out, we'll still be trying to send the SMS 3 times?

rudigiesler: Yes, the worker will attempt to send up to 3 times (for any HTTP errors), and then send a response to the user. So it is possible that if a timeout happens here, we'll send the SMS, and then try to respond to the user's USSD session, but fail because the session has since timed out.

These retries are immediate though, so hopefully this is a short period of time. But there is a small chance that the timeout will happen while we're doing this, yes.

PR #314, opened 2020-06-11
Fix msisdn and fields on PMTCT flow start

erikh360:
We need to start the flow on the msisdn entered if it was entered.
We need to send registered_by, mha and swt when starting the flow.


OMosimege (commented):
@erikh360 the mha and swt aren't automatically set in rapidpro?

Otherwise, I think it look great 👍

erikh360:
@OMosimege it does have defaults in the flow, but that means all of them will go through as WA, even SMS contacts.

PR #317, opened 2020-07-06
educational info flow stubs

moh-moola:


src/ussd_higherhealth_healthcheck.js
```diff
@@ -224,6 +225,714 @@ go.app = (function () {
...
+                "Parow",
+                "Giyani",
+                "Makhado",
+                "Polokwane",
+                "Kimberley",
+                "Mafikeng",
+                "Potchefstroom",
+                "Rustenburg",
+                "Mbombela",
+                "Middelburg"
```
rudigiesler: I don't think that this data structure will work.

An institution like UNISA has campuses all over the country. If the user selected "Western Cape", then we only want to display the western cape campuses, not all the campuses across the country. This list is way too big to be going through page-by-page on USSD.

I would suggest something like:
```
{
  "ZA-WC": {
    "UNISA": [
      "George",
      "Parow"
    ],
    ...
  },
  ...
}

File: src/ussd_higherhealth_healthcheck.js [hunk 2]
```diff
@@ -224,6 +225,714 @@ go.app = (function () {
...
+      return new PaginatedChoiceState(name, {
+        question: $([
+          "Select your university.",
+          "",
+          "Reply:"
+        ].join("\n")),
+        accept_labels: true,
+        options_per_page: null,
+        choices: choices,
+        next: "state_campus"
```
rudigiesler: I see we haven't defined the `more` and `back` labels.

I would define them, even if we set them to the default "Back" and "Next" values, so that when the content gets pulled for translation, the "Back" and "Next" also gets pulled and sent to the translators.

src/ussd_higherhealth_healthcheck.js
```diff
@@ -224,6 +225,714 @@ go.app = (function () {
...
+            'Boland', 'Cape Peninsula University of Technology (CPUT)', 'College of Cape Town', 'False Bay', 'Nelson Mandela University (NMU)', 'Northlink', 'South Cape', 'Stellenbosch University (SU)', 'UNISA', 'University of Cape Town (UCT)', 'University of Western Cape (UWC)', 'West Coast']
+      };
+      var selected_prov = self.im.user.answers.state_province;
+      var choices = [];
+      for(var i=0; i<universities_by_province[selected_prov].length; i++){
+        choices[i] = new Choice(
+            universities_by_province[selected_prov][i],
+            $(universities_by_province[selected_prov][i])
+        );
+      }
```
rudigiesler, line 269: I see we have an "Other" university choice for Campus, but no other choice here in the lists, so the users have no way of selecting that.

We should probably add "Other" to the list of universities.

I would add it at the end of `choices`, instead of adding it manually to each province's list.

src/ussd_higherhealth_healthcheck.js
```diff
@@ -224,6 +225,714 @@ go.app = (function () {
...
+                "Carletonville",
+                "Central Office",
+                "Krugersdorp",
+                "Krugersdorp West",
+                "Randfontein"
+
+        ],
+            "Other":[
+                "Other"
+            ],
```
rudigiesler: We have "Other" here, but what if the user selects a University, but wants to select an "Other" for campus, because their campus isn't on the list?

I think we should also have "Other" in all the lists of campuses. Probably added to the `choices` list, instead of manually adding it to each of these lists.

We probably also want some testing for the "Other" logic.

test/ussd_higherhealth_healthcheck.test.js
```diff
@@ -373,6 +376,9 @@ describe("ussd_higherhealth_healthcheck app", function () {
         it("should display an error on invalid input", function () {
             return tester
                 .setup.user.state("state_age")
+                .setup.user.answers({
+                    state_province: 'ZA-GT',
+                })
```
rudigiesler: I'm not sure why we're setting up `state_province` here? We haven't made any changes to `state_age` for it to require the province. I think we only need `state_province` on the `should go to state_university` test, since then it's going to `state_university` where it needs the province.

src/ussd_higherhealth_healthcheck.js
```diff
@@ -224,10 +225,767 @@ go.app = (function () {
...
+                "Krugersdorp West",
+                "Randfontein"
+
+        ],
+        };
+      var selected_prov = self.im.user.answers.state_province;
+      var selected_uni = self.im.user.answers.state_university;
+      var choices = [];
+      var campus_list = [];
+      if (selected_uni !== 'UNISA') campus_list = campuses_by_university[selected_uni];
```
rudigiesler: I think there's a misunderstanding here. It's not just UNISA that has campuses in multiple provinces, that was just the example I chose. We need to do this for all campuses, not just UNISA campuses.

src/ussd_higherhealth_healthcheck.js
```diff
@@ -224,10 +225,767 @@ go.app = (function () {
...
+      return new PaginatedChoiceState(name, {
+        question: $([
+          "Select your university.",
+          "",
+          "Reply:"
+        ].join("\n")),
+        accept_labels: true,
+        options_per_page: null,
+        more: "More",
+        back: "Back",
```
rudigiesler: These labels aren't surrounded in `$(...)`, so they won't be extracted for translation, nor will they be translated.

src/ussd_higherhealth_healthcheck.js
```diff
@@ -224,10 +225,767 @@ go.app = (function () {
...
+      return new PaginatedChoiceState(name, {
+        question: $([
+          "Select your campus.",
+          "",
+          "Reply:"
+        ].join("\n")),
+        accept_labels: true,
+        options_per_page: null,
+        more: "More",
+        back: "Back",
```
rudigiesler: These labels also need to be surrounded by `$(...)`

test/ussd_higherhealth_healthcheck.test.js
```diff
@@ -389,14 +392,227 @@ describe("ussd_higherhealth_healthcheck app", function () {
...
+                .setup.user.answers({
+                    state_province: 'ZA-GT'
+                })
+                .input("A")
+                .check.interaction({
+                    state: "state_university",
+                    reply: [
+                        "Select your university.",
+                        "",
+                        "Reply:",
```
rudigiesler, line 443: I see for invalid input, we're using the same text as we do for the initial question. Is that intended?

rudigiesler (commented):
We're collecting the university and campus, but we're not saving/submitting that information anywhere. Is it intended that we'll do that in a separate PR?

src/ussd_higherhealth_healthcheck.js
moh-moola: 👍

test/ussd_higherhealth_healthcheck.test.js
moh-moola: yes

rudigiesler (approved):

PR #319, opened 2020-07-09
Feature/momza 1380 ussd drop off monitoring

OMosimege:
In this PR, we add metrics to the clinic line to check for
1. number of users that start a registration
2. tracking users that enter each state

### DIFF
--- a/src/ussd_clinic_rapidpro.js
+++ b/src/ussd_clinic_rapidpro.js
@@ -12,6 +12,7 @@ go.app = function() {
     var MenuState = vumigo.states.MenuState;
     var PaginatedState = vumigo.states.PaginatedState;
     var PaginatedChoiceState = vumigo.states.PaginatedChoiceState;
+    var MetricsHelper = require('go-jsbox-metrics-helper');
 
 
     var GoNDOH = App.extend(function(self) {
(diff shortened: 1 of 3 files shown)


src/ussd_clinic_rapidpro.js
```diff
@@ -35,6 +36,20 @@ go.app = function() {
...
+            self.metric_prefix = [self.env, self.im.config.name].join('.');
+            
+            self.im.on('state:enter', function(e) {
+                return self.im.metrics.fire.sum('enter.' + e.state.name, 1);
+            });
+
+            var mh = new MetricsHelper(self.im);
+            mh
+                // Total sum of users for each state for app
+                // This adds <env>.ussd_clinic_rapidpro.sum.users_per_state.transient 'sum' metric
```
rudigiesler: I don't think that this adds a `<env>.ussd_clinic_rapidpro.sum.users_per_state.transient` metric, I think it adds a `<env>.ussd_clinic_rapidpro.sum.unique_users` `last` metric, and a `<env>.ussd_clinic_rapidpro.sum.unique_users.transient` `sum` metric

rudigiesler (commented):
Just one small comment, looks good otherwise :)

rudigiesler (approved):

PR #320, opened 2020-07-14
add language choice screen

moh-moola:
Add a screen to change a users language then continue to the terms screen

### DIFF
--- a/go-app-ussd_higherhealth_healthcheck.js
+++ b/go-app-ussd_higherhealth_healthcheck.js
@@ -388,7 +388,7 @@ go.app = (function () {
       for(var i=0; i<universities_by_province[selected_prov].length; i++){
         choices[i] = new Choice(
             universities_by_province[selected_prov][i],
-            $(universities_by_province[selected_prov][i])
+            $("{{ label}}").context({label: universities_by_province[selected_prov][i]})
         );
       }
       choices[choices.length] = new Choice('Other', $("Other"));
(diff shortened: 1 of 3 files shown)


src/ussd_higherhealth_healthcheck.js
```diff
@@ -78,18 +79,38 @@ go.app = (function () {
...
+        choices: [
+          new Choice("af", $("Afrikaans")),
+          new Choice("en", $("English")),
+          new Choice("st", $("Sotho")),
+          new Choice("xh", $("Xhosa")),
+          new Choice("zu", $("Zulu")),
+        ],
+        next: "state_terms"
+      });
+    });
```
rudigiesler, line 112: I think this should be the start screen, so the user selects their language first, and then the rest of the interaction is in that language.

go-app-ussd_higherhealth_healthcheck.js
```diff
@@ -181,18 +182,38 @@ go.app = (function () {
...
+          "",
+          "Reply:"
+        ].join("\n")),
+        accept_labels: true,
+        choices: [
+          new Choice("af", $("Afrikaans")),
+          new Choice("en", $("English")),
+          new Choice("st", $("Sotho")),
+          new Choice("xh", $("Xhosa")),
+          new Choice("zu", $("Zulu")),
```
rudigiesler: Can we use ISO639-3 codes here, instead of ISO639-1. Even though the current languages fit into ISO639-1, not all official South African languages do, so if we ever expand this to all South African languages, then the two letter codes will no longer be sufficient.

test/ussd_higherhealth_healthcheck.test.js
```diff
@@ -94,7 +96,44 @@ describe("ussd_higherhealth_healthcheck app", function () {
...
+                    char_limit: 140
+                })
+                .run();
+        });
+
+        it("should go to change the language", function () {
+            return tester
+                .setup.user.state("state_language")
+                .input("2")
+                .check.user.state("state_terms")
```
rudigiesler, line 132: We should also check that the user's language has changed.

src/ussd_higherhealth_healthcheck.js
moh-moola: Then that means it would be mandatory for a user to always take that step, it would make sense if it is option, with English  being the default

PR #321, opened 2020-07-14
add education info to api

moh-moola:
Add educational information to the API


src/ussd_higherhealth_healthcheck.js
```diff
@@ -1150,6 +1150,10 @@ go.app = (function () {
           province: answers.state_province,
           city: answers.state_city,
           age: answers.state_age,
+          university: answers.state_university,
+          university_other: answers.state_university_other,
+          campus: answers.state_campus,
+          campus_other: answers.state_campus_other,
```
rudigiesler: These aren't valid fields in the API, so this request will fail.

I think we'll need to put this in the `data` field, see https://github.com/praekeltfoundation/higher-health-django/blob/develop/higher_health/tasks.py#L63-L88

moh-moola: You mean that the API endpoint isn't expecting these values?

rudigiesler: Yes, those aren't fields in the API spec, because they're specific to HigherHealth, they're not applicable to a generic HealthCheck.

moh-moola: @rudigiesler addressed issue on https://github.com/praekeltfoundation/higher-health-django/pull/109

rudigiesler: I think there's a misunderstanding here. That PR doesn't solve this issue at all.
The code in `higher-health-django` is just an example of what a request to the API could look like, that's not where the API is defined at all.
And we don't want to be changing the API definition for this, the university and campus aren't generic enough to add to the API. We should instead be placing this data in the `data` field, similar to how it is done in the task in `higher-health-django`

moh-moola: @rudigiesler I have since updated the api call and amended the tests

rudigiesler: I still don't think that this is quite right. I think the additional data should be in the `data` field, since these aren't fields that are present in the API, the API will return a 400 error status if we attempt this. But there's no validation on the `data` field, as long as it's valid JSON. ie:
```
city: answers.state_city,
age: answers.state_age,
data: {
  university: {
    name: answers.state_university
  },
  ...
}
```

src/ussd_higherhealth_healthcheck.js
```diff
@@ -402,6 +412,7 @@ go.app = (function () {
         opts.http_error_count = _.get(opts, "http_error_count", 0) + 1;
         if (opts.http_error_count === 3) {
           self.im.log.error(e.message);
+          console.log(e, e.message, '===========================');
```
rudigiesler: This should be removed, `console` is not accessible while in the sandboxed environment, so the app will crash if it reaches this point

rudigiesler (approved):

PR #326, opened 2020-07-29
Remove unused applications

rudigiesler:
This PR does some cleanup by removing all code related to applications that are no longer in use


erikh360 (approved):
Awesome!

PR #331, opened 2020-08-14
HealthCheck confirmed contact

rudigiesler:
This adds a confirmed contact lookup in the beginning of the flow, and then gives a slightly different user flow if that user is a confirmed contact.

Content is in this sheet: https://docs.google.com/spreadsheets/d/1cVjPb-ae1qOh4wxJ2skm0kS4QrUoCp9opm9Pqd59VGc/edit#gid=222340857

### DIFF
--- a/go-app-ussd_covid19_triage.js
+++ b/go-app-ussd_covid19_triage.js
@@ -12,6 +12,7 @@ go.app = (function () {
   var FreeText = vumigo.states.FreeText;
   var ChoiceState = vumigo.states.ChoiceState;
   var utils = require("seed-jsbox-utils").utils;
+  var crypto = require("crypto");
 
 
   var GoNDOH = App.extend(function (self) {
@@ -106,6 +107,7 @@ go.app = (function () {
     });
 
     self.states.add("state_get_confirmed_contact", function(name, opts) {
+      self.im.user.answers.google_session_token = crypto.randomBytes(20).toString("hex");
       var msisdn = utils.normalize_msisdn(self.im.user.addr, "ZA");
       var whatsapp_id = _.trimStart(msisdn, "+");
       return new JsonApi(self.im).get(
@@ -267,7 +269,7 @@ go.app = (function () {
     });
 
     self.add("state_city", function (name) {
-      if(self.im.user.answers.state_city) {
+      if(self.im.user.answers.state_city && self.im.user.answers.city_location) {
         return self.states.create("state_age");
       }
       var question = $(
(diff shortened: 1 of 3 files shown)


test/ussd_covid19_triage.test.js
```diff
@@ -152,6 +185,41 @@ describe("ussd_covid19_triage app", function () {
...
+                .check.interaction({
+                    state: "state_welcome",
+                    reply: [
+                        "The Dept of Health: you have been in contact with someone who has " +
+                        "COVID-19. Isolate for 10 days & answer these questions.",
+                        "",
+                        "Reply",
+                        "1. START"
+                    ].join("\n"),
+                    char_limit: 140
```
moh-moola, line 202: not sure why you have a limit of 140 chars' where in all other places its 160

test/ussd_covid19_triage.test.js
```diff
@@ -687,6 +1155,45 @@ describe("ussd_covid19_triage app", function () {
...
+        });
+        it("should display the error on invalid input confirmed contact", function () {
+            return tester
+                .setup.user.state("state_tracing")
+                .setup.user.answer("confirmed_contact", true)
+                .input("A")
+                .check.interaction({
+                    state: "state_tracing",
+                    reply: [
+                        "Please use numbers from the list.",
```
moh-moola, line 1184: shouldn't this error be a standard text? There are some places where "`This service works best when you select numbers from the list.`" is used... should that be the case?

src/ussd_covid19_triage.js
```diff
@@ -399,20 +713,27 @@ go.app = (function () {
...
           source: "USSD",
           province: answers.state_province,
           city: answers.state_city,
+          city_location: answers.city_location,
           age: answers.state_age,
           fever: answers.state_fever,
           cough: answers.state_cough,
           sore_throat: answers.state_sore_throat,
           difficulty_breathing: answers.state_breathing,
+          smell: answers.state_taste,
```
moh-moola: In the long run this will be confusing especially for someone touching the code base for the 1st time

src/ussd_covid19_triage.js
```diff
@@ -255,6 +428,36 @@ go.app = (function () {
...
+          } else if (age < 40) {
+            self.im.user.answers.state_age = "18-40";
+            return;
+          } else if (age <= 65) {
+            self.im.user.answers.state_age = "40-65";
+            return;
+          } else if (age < 150) {
+            self.im.user.answers.state_age = "40-65";
+            return;
+          }
```
moh-moola: You could use a switch statement here

test/ussd_covid19_triage.test.js
rudigiesler: It's the first screen of the session. Some MNOs have a lower limit for the first screen you show a user vs the rest of the screens.

test/ussd_covid19_triage.test.js
rudigiesler: There's not enough characters to be able to show the full error text, so for certain screens it was shortened. The error text is also present in the content doc

src/ussd_covid19_triage.js
rudigiesler: We can't change the name of the API field, but I can see if we can change the name of the state.

src/ussd_covid19_triage.js
rudigiesler: I'm not sure how I could use a switch statement without it being very verbose? Switch statements have to match on exact results, so I'm not sure how to do it besides doing something like:
```
switch(age) {
  case 1:
  case 2:
  case 3:
  case 4:
  ...
    self.im.user.answers.state_age = "<18";
    break;

  case 18:
  case 19:
  ...
}
```

moh-moola:
```
switch(age) {
  case age < 18:
      self.im.user.answers.state_age = "<18";
      break;
  ...
}
```

rudigiesler: Yeah that doesn't work for me:
```
> var age = 23;
> var result = null;
> switch(age) {
... case age < 18:
...   result = "<18";
...   break;
... case age < 40:
...   result = "18-40";
...   break;
... }
> result
null
```
And I don't see that being possible anywhere in the docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

moh-moola: https://www.w3schools.com/js/tryit.asp?filename=tryjs_switch, conditional case statements are acceptable

rudigiesler: That link is still a `case 0:` `case 1:`, etc. I don't see any `case age < 18:`, and it doesn't work for me when I try it

moh-moola: @rudigiesler not a train smash, but this does make sense in a switch statement

go-app-ussd_covid19_triage.js
```diff
@@ -258,6 +431,36 @@ go.app = (function () {
...
+            self.im.user.answers.state_age = "<18";
+            return;
+          } else if (age < 40) {
+            self.im.user.answers.state_age = "18-40";
+            return;
+          } else if (age <= 65) {
+            self.im.user.answers.state_age = "40-65";
+            return;
+          } else if (age < 150) {
+            self.im.user.answers.state_age = "40-65";
```
miltontony: This should be `>65`

go-app-ussd_covid19_triage.js
```diff
@@ -258,6 +431,36 @@ go.app = (function () {
...
+          } else if (age < 18) {
+            self.im.user.answers.state_age = "<18";
+            return;
+          } else if (age < 40) {
+            self.im.user.answers.state_age = "18-40";
+            return;
+          } else if (age <= 65) {
+            self.im.user.answers.state_age = "40-65";
+            return;
+          } else if (age < 150) {
```
miltontony, line 454: Why not just > 65?

go-app-ussd_covid19_triage.js
```diff
@@ -364,31 +632,77 @@ go.app = (function () {
...
       });
     });
 
-    self.add("state_tracing", function (name) {
+    self.add("state_exposure", function (name) {
       return new ChoiceState(name, {
         question: $([
-          "Please confirm that the information you shared is correct & that the National " +
-          "Department of Health can contact you if necessary?",
+          "Have you been in close contact to someone confirmed to be infected with COVID19?",
```
miltontony, line 642: @rudigiesler should this question not be skipped for the confirmed contact case?

go-app-ussd_covid19_triage.js
rudigiesler: I wanted to avoid bad data being accidentally added. If they accidentally type an extra number or something, eg. they type `566` instead of `56`

go-app-ussd_covid19_triage.js
rudigiesler: Yes it is. For confirmed contacts, the state before this one is `state_breathing`. If you look at `state_breathing`, you'll see for confirmed contacts, we set `state_exposure` to `yes`, and go to `state_taste_and_smell`. So confirmed contacts won't see `state_exposure`, and will be submitted with `state_exposure=yes`

go-app-ussd_covid19_triage.js
miltontony: +1

miltontony (approved):

PR #347, opened 2020-09-17
Change welcome, remove language, and disable SMS registration

rudigiesler:
This PR changes the welcome message, removes language selection defaulting to English, and shows an error instead of allowing users without a WhatsApp account to register.

This is all for the switching off of SMS for MomConnect. This does not address any changes required for the POPI line, those will be done in a separate PR


OMosimege (approved):
👍

PR #349, opened 2020-09-18
Don't allow switching back to SMS

rudigiesler:
There are 2 parts to this
1. We don't give the option to switch channels if they're on WhatsApp
2. If they change numbers, then the target number must be on WhatsApp


OMosimege (approved):
👍

PR #359, opened 2021-02-13
Revert institution updates and add institutions

rudigiesler:
The previous institution update was too many institutions, so we're reverting that and just adding what is required


miltontony:
👍

PR #361, opened 2021-03-05
Add webhook call to save healthcheck start

erikh360:


test/ussd_covid19_triage.test.js
```diff
@@ -86,6 +102,7 @@ describe("ussd_covid19_triage app", function () {
                         }
                     });
                 })
+                .inputs({ session_event: "new", to_addr: "*123#" })
```
rudigiesler: This can probably be `.input(`, since it's only a single message

test/ussd_covid19_triage.test.js
```diff
@@ -125,6 +158,7 @@ describe("ussd_covid19_triage app", function () {
...
```
rudigiesler: Same here

rudigiesler (commented):
Just one minor comment around input/inputs, otherwise looks good

rudigiesler (approved):

PR #366, opened 2021-03-18
Extra screens for supporter change flows

DevChima:


go-app-ussd_mcgcc_rapidpro.js
```diff
@@ -980,6 +982,183 @@ go.app = function() {
...
+                error: $("Please try again. Reply with the nr that matchs your answer, e.g. 1").context(),
+                accept_labels: true,
+                choices: [
+                    new Choice("state_supporter_change_msisdn_rapidpro", $("Yes")),
+                    new Choice("state_supporter_new_msisdn", $("No, I want to retype my number"))
+                ]
+            });
+        });
+              
+        /*self.add("state_supporter_new_msisdn_whatsapp_contact_check", function(name, opts) {
```
erikh360: Should this be commented out?

erikh360 (commented):
Looks good  but I think your go-* file needs to be regenerated.

go-app-ussd_mcgcc_rapidpro.js
DevChima: Yeah, it's older than the one I currently have locally. I'll regenerate

erikh360 (approved):

PR #367, opened 2021-03-18
added npm grunt-pot, as jspot is no longer supported

OMosimege:
In this PR, I have added `grunt-pot`, as `jspot` is no longer supported

see this https://www.npmjs.com/package/grunt-pot for reference


rudigiesler:
jspot is one of our tools: https://github.com/praekeltfoundation/jspot , what about `jspot` is no longer working?

Other than that, we probably want a PO file per app, not a single file that contains the translation strings for all apps combined.

PR #373, opened 2021-03-25
Feature/translations upload

OMosimege:
In this PR, is more content upload

Plus setting the `lang` in state_language


src/ussd_tb_check.js
```diff
@@ -14,6 +14,14 @@ go.app = (function () {
     App.call(self, "state_start");
     var $ = self.$;
 
+    self.languages = {
+      zul: "isiZulu",
+      xho: "isiXhosa",
+      afr: "Afrikaans",
+      eng: "English",
+      sot: "Sesotho"
+  };
```
rudigiesler: It doesn't seem like we're using this anywhere?

OMosimege: ahh, I thought this was needed for the testing setup. Will remove

rudigiesler (commented):
👍

PR #384, opened 2021-04-06
add sponsor message for hh

codiebeulaine:


CourtneyMoxley:
On line 674, please change text to "HIGHER HEALTH supported by Lifebuoy, European Union and HWESTA"

rudigiesler (approved):

PR #398, opened 2021-04-27
update institutions hh

codiebeulaine:


rudigiesler (approved):
Looks good. Just want to double check that we want to be removing some of the institutions

PR #400, opened 2021-05-11
add hh heealthcheck receipt

codiebeulaine:


src/ussd_higherhealth_healthcheck.js
```diff
@@ -121,14 +120,22 @@ go.app = (function () {
...
-          "",
-          "Reply"
-        ].join("\n"));
+        return new MenuState(name, {
+          question: question,
+          error: $("This service works best when you select numbers from the list"),
+          accept_labels: true,
+          choices: [
+            new Choice("state_terms", $("START")),
+            new Choice("state_display_risk", $("RECEIPT"))
```
rudigiesler, line 175: in `state_start`, all the answers are replaced with what is found in the user's profile, if they have one. That means that we won't have the information required to be able to display a receipt. 

Also, the `answers` data is only kept around for an hour, so if we need to be able to show a receipt for longer than that, we can't rely on the answers data.

We probably need to add to the current healthcheck API to allow fetching the healthchecks (or last healthcheck) for a user (making sure that it's indexed well and performant), and then use that API to fetch their last healthcheck, making sure that it's within some timeframe (eg. you should only be able to view your receipt for 24 hours, or more likely, you can only view the receipt on the same day that you did the healthcheck). We would also probably not want to show the `RECEIPT` option if they haven't completed a healthcheck within the given timeframe.

.jshintrc
```diff
@@ -10,5 +10,6 @@
   "expr": true,
   "node": true,
   "undef": true,
-  "unused": "vars"
+  "unused": "vars",
+  "esversion": 6
```
rudigiesler: Nodejs 4 only supports 43% of ES6: https://node.green/

src/ussd_higherhealth_healthcheck.js
```diff
@@ -18,6 +18,7 @@ go.app = (function () {
     var $ = self.$;
 
     self.calculate_risk = function () {
+      // if has receipt for today, send that receipt's risk
```
rudigiesler: This comment seems out of place

src/ussd_higherhealth_healthcheck.js
```diff
@@ -110,32 +111,73 @@ go.app = (function () {
...
     });
-
     self.states.add("state_welcome", function(name) {
       self.im.user.answers.google_session_token = crypto.randomBytes(20).toString("hex");
       var question;
-      if (self.im.user.answers.returning_user) {
+      var today = new Date();
+      var date = today.getFullYear()+'-'+(today.getMonth()+1)+'-'+today.getDate();
+      var msisdn = self.im.user.addr;
+      new JsonApi(self.im).get(
```
rudigiesler: We probably want to return this promise

test/ussd_higherhealth_healthcheck.test.js
```diff
@@ -158,7 +158,8 @@ describe("ussd_higherhealth_healthcheck app", function () {
                         "No result SMS will be sent. Continue or WhatsApp HI to 0600110000",
                         "",
                         "Reply",
-                        "1. START"
+                        "1. START",
+                        "2. RECEIPT"
                     ].join("\n"),
                     char_limit: 140
                 })
```
rudigiesler, line 264: We'll want tests for the new functionality..

go-app-ussd_higherhealth_healthcheck.js
```diff
@@ -1701,32 +1702,81 @@ go.app = (function () {
...
+      return new JsonApi(self.im).get(
+        // use dictionary params
+        self.im.config.eventstore.url + "/api/v3/covid19triage/", {
+        headers: {
+          "Authorization": ["Token " + self.im.config.eventstore.token],
+          "User-Agent": ["Jsbox/HH-Covid19-Triage-USSD"]
+        },
+        params: {
+          "timestamp_gt": date + "T00:00:00+02:00",
+          "msisdn": msisdn
```
rudigiesler, line 1719: Do we need to make sure that this msisdn is normalised? eg `+27820001001` instead of `27820001001`?

go-app-ussd_higherhealth_healthcheck.js
```diff
@@ -1701,32 +1702,81 @@ go.app = (function () {
...
+          choices: [
+            new Choice("state_terms", $("START")),
+            new Choice("state_display_risk", $("RECEIPT"))
+          ]
+        });
+      }, function (e) {
+        if(_.get(e, "response.code") === 404) {
+          self.im.user.answers = {returning_user: false};
+          return self.states.create("state_welcome");
+        }
```
rudigiesler: I don't think that this request will ever return a 404, it will just return a 200 with an empty list

go-app-ussd_higherhealth_healthcheck.js
```diff
@@ -1701,32 +1702,81 @@ go.app = (function () {
...
+            accept_labels: true,
+            choices: [
+              new Choice("state_terms", $("START"))
+            ]
+          });
+        }
+        var last_result = results[results.length - 1];
+        self.im.user.answers.risk = last_result.risk;
+        self.im.user.answers.state_first_name = last_result.first_name;
+        self.im.user.answers.state_last_name = last_result.last_name;
```
rudigiesler, line 1752: Looking at `state_display_risk`, it seems we also need the `state_tracing` answer

rudigiesler (approved):

PR #413, opened 2021-10-18
skip location/city for minors

DevChima:


go-app-ussd_tb_check.js
```diff
@@ -725,35 +728,43 @@ go.app = function () {
     self.add("state_submit_data", function (name, opts) {
       var answers = self.im.user.answers;
       var msisdn = utils.normalize_msisdn(self.im.user.addr, "ZA");
+      var payload = {
+        data: {
+          msisdn: msisdn,
+          source: "USSD",
+          language: answers.state_language,
+          province: answers.state_province,
+          //city: answers.state_city,
```
erikh360: Should these be commented out? We can probably just remove

erikh360 (approved):
looks good, just the one comment

go-app-ussd_tb_check.js
DevChima: I needed to run another test after removing that line in code. I'll correct.

erikh360 (approved):
🍰

PR #417, opened 2021-11-09
Check that supporter's new msisdn is on Whatsapp

DevChima:


erikh360 (commented):
Looks good, can we also create a test for the new `state_supporter_new_msisdn_whatsapp_contact_check` state

erikh360 (approved):

PR #424, opened 2021-11-24
Changes to the USSD clinic registration flow

DevChima:
This pr moves the clinic code question up before the subscriptions check

### DIFF
--- a/src/ussd_clinic_rapidpro.js
+++ b/src/ussd_clinic_rapidpro.js
@@ -140,16 +140,31 @@ go.app = function() {
                         );
                     }
                 },
-                next: "state_clinic_code"
+                next: "state_whatsapp_check"
             });
+        });
+
+        self.add("state_whatsapp_contact_check", function(name, opts) {
+            var msisdn = utils.normalize_msisdn(
+                _.get(self.im.user.answers, "state_enter_msisdn", self.im.user.addr), "ZA");
+            self.whatsapp.contact_check(msisdn, false).then(_.noop, _.noop);
+            return self.whatsapp.contact_check(msisdn, true)
+                .then(function(result) {
+                    self.im.user.set_answer("on_whatsapp", result);
+                    return self.states.create("state_clinic_code");
+                }).catch(function(e) {
+                    // Go to error state after 3 failed HTTP requests
+                    opts.http_error_count = _.get(opts, "http_error_count", 0) + 1;
+                    if(opts.http_error_count === 3) {
+                        self.im.log.error(e.message);
+                        return self.states.create("__error__", {return_state: name});
+                    }
+                    return self.states.create(name, opts);
+                });
         });
 
         self.add("state_clinic_code", function(name, opts) {
             var text;
-            // Run a no-wait contact check in the background to populate the cache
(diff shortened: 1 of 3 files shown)


src/ussd_clinic_rapidpro.js
```diff
@@ -140,16 +140,62 @@ go.app = function() {
...
 
+        self.add("state_clinic_code", function(name, opts) {
+            var text;
+            // Run a no-wait contact check in the background to populate the cache
+            var msisdn = utils.normalize_msisdn(
+                _.get(self.im.user.answers, "state_enter_msisdn", self.im.user.addr), "ZA");
+            self.whatsapp.contact_check(msisdn, false).then(_.noop, _.noop);
+            if(!self.im.user.answers.on_whatsapp) {
+                return self.states.create("state_not_on_whatsapp");
+            }
```
rudigiesler, line 169: I don't see anywhere that would set this `on_whatsapp` answer at this point, it only gets set later in the flow, so this condition will always trigger, and the use will always be sent to `state_not_on_whatsapp`

src/ussd_clinic_rapidpro.js
```diff
@@ -140,16 +140,77 @@ go.app = function() {
...
                 },
-                next: "state_get_contact"
+                next: "state_whatsapp_check"
             });
         });
 
+        self.add("state_whatsapp_contact_check", function(name, opts) {
+            var msisdn = utils.normalize_msisdn(
+                _.get(self.im.user.answers, "state_enter_msisdn", self.im.user.addr), "ZA");
+            self.whatsapp.contact_check(msisdn, false).then(_.noop, _.noop);
```
rudigiesler: If we're immediately doing a wait contact check after this, then this nowait check is not necessary. This is only an advantage if we can do it earlier on, so that when we do the wait check it returns quicker.

src/ussd_clinic_rapidpro.js
```diff
@@ -140,16 +140,77 @@ go.app = function() {
...
+                        self.im.log.error(e.message);
+                        return self.states.create("__error__", {return_state: name});
+                    }
+                    return self.states.create(name, opts);
+                });
+        });
+
+        self.add("state_clinic_code", function(name, opts) {
+            var text;
+            if(!self.im.user.answers.on_whatsapp) {
```
rudigiesler, line 167: In `state_start`, if they choose the `yes` option, they go straight here without going through `state_whatsapp_contact_check`, so this field won't be set, and they'll always end up on `state_not_on_whatsapp`

src/ussd_clinic_rapidpro.js
```diff
@@ -704,32 +724,14 @@ go.app = function() {
...
-                    return self.states.create(name, opts);
-                });
-        });
-
         self.add("state_start_popi_flow", function(name, opts) {
-            if(!self.im.user.answers.on_whatsapp) {
+            /*if(!self.im.user.answers.on_whatsapp) {
                 return self.states.create("state_not_on_whatsapp");
-            }
+            }*/
```
rudigiesler: We should probably remove this check, instead of commenting it out.

rudigiesler (approved):

PR #426, opened 2021-12-01
dont ask name or surname if <18

codiebeulaine:


erikh360 (approved):
Looks good but two suggestions:
Make sure the api accepts empty values for those fields
A test to confirm it builds the request body properly with those fields not set

codiebeulaine:
@erikh360 it is not required here
https://github.com/praekeltfoundation/ndoh-hub/blob/develop/eventstore/models.py#L459

erikh360 (approved):

PR #428, opened 2021-12-02
Feature mcgcc partner messaging

DevChima:
This pr has two commits. 
The first one removes the "change channel" option from the menu but we realized that we need to keep it in to allow contacts on SMS switch over to Whatsapp. 

The second pr replaces the "change channel" option but makes sure it's only visible to contacts on SMS. 
Contacts on Whatsapp never get to see it.


erikh360 (commented):
just the one comment, I think we can simplify this

rudigiesler (approved):

PR #431, opened 2022-01-26
Clinic Registration Update

DevChima:
Miro flow: https://miro.com/app/board/o9J_lo5HSr8=/

Passport holders should be allowed to enter their age
We need to check if contact is underage regardless of their means of identification
We also need to send them  a custom message if they are underage doing self registration or an underage registree

### DIFF
--- a/go-app-ussd_clinic_rapidpro.js
+++ b/go-app-ussd_clinic_rapidpro.js
@@ -997,10 +997,10 @@ go.app = function() {
                 msisdn = utils.normalize_msisdn(self.im.user.get_answer("state_enter_msisdn"), "ZA");
                 self_registration = false;
             }
-            if (((typeof self.im.user.get_answer("state_sa_id_no")!= "undefined") && (_.toUpper(self.im.user.get_answer("state_basic_healthcare")) != "CONFIRM"))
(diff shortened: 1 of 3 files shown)


go-app-ussd_clinic_rapidpro.js
```diff
@@ -966,9 +985,84 @@ go.app = function() {
...
+            var msisdn, self_registration, age;
+            if (typeof self.im.user.get_answer("state_enter_msisdn") === "undefined"){
+                msisdn = utils.normalize_msisdn(self.im.user.addr, "ZA");
+                self_registration = true;
+            }
+            else{
+                msisdn = utils.normalize_msisdn(self.im.user.get_answer("state_enter_msisdn"), "ZA");
+                self_registration = false;
+            }
+            if (((typeof self.im.user.get_answer("state_sa_id_no")!= "undefined") && (_.toUpper(self.im.user.get_answer("state_basic_healthcare")) != "CONFIRM"))
```
erikh360: Wouldn't this be easier to check `state_id_type`?

erikh360 (commented):
One comment for now, applicable to all 3 type checks

PR #433, opened 2022-02-03
End screen for mothers on active prebirth registration

DevChima:
We want to detect that they have an active subscription and then show a screen with instructions on how to end existing subscription.

### DIFF
--- a/src/ussd_clinic_rapidpro.js
+++ b/src/ussd_clinic_rapidpro.js
@@ -310,9 +310,7 @@ go.app = function() {
         self.add("state_active_subscription_2", function(name) {
             var choices = [];
             var contact = self.im.user.answers.contact;
-            if (!self.contact_edd(contact)) {
-                choices.push(new Choice("state_edd_month", $("Register a new pregnancy")));
-            }
+            choices.push(new Choice("state_edd_month", $("Register a new pregnancy")));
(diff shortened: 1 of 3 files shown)


src/ussd_clinic_rapidpro.js
```diff
@@ -403,6 +401,9 @@ go.app = function() {
         });
 
         self.add("state_edd_month", function(name) {
+            if (typeof self.im.user.get_answer("state_active_subscription") != "undefined") {
```
erikh360: This check is a bit hard to read and there is a helper function for this that was used here: https://github.com/praekeltfoundation/ndoh-jsbox/pull/433/files#diff-5c4090be76401c5461c39da6838993894732990276768da57f543ead9240f2baL313

src/ussd_clinic_rapidpro.js
```diff
@@ -438,6 +439,20 @@ go.app = function() {
             });
         });
 
+        self.states.add("state_active_prebirth_end", function(name) {
+            var edd = _.get(self.im.user.get_answer("contact"), "fields.edd");
```
erikh360: Please use the helper function here too.

erikh360 (commented):
Two comments to fix

erikh360 (approved):

PR #434, opened 2022-02-09
Prebirth <42 weeks should start the prebirth flow and edd value shoul…

DevChima:
…d be displayed correctly

Because a new prebirth registration skips state_message_type if EDD > 42 weeks, contacts in this category enter details for pre-birth but are started on the post-birth flow.

Secondly, the helper function ony returns EDD if EDD < 42 weeks so if greater and a contact has the condition above we send a message with an empty EDD such as "The number 0827081992 is already receiving messages from MomConnect for....."

### DIFF
--- a/go-app-ussd_clinic_rapidpro.js
+++ b/go-app-ussd_clinic_rapidpro.js
@@ -508,7 +508,7 @@ go.app = function() {
             };
 
             var subscriptions = [];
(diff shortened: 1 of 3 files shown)


erikh360 (approved):

Buhle79 (commented):
+1

PR #445, opened 2022-03-02
Add MQR FAQ app

erikh360:
Sorry about the size, we an ignore the go-* file since thats just generated.


DevChima:
A LOT but looks good. :)

PR #446, opened 2022-03-10
Update for different content format

erikh360:


DevChima:
I'm not familiar with the flows so it's a bit hard to confirm that no screens were skipped but code looks good.

Buhle79 (commented):
+1

PR #454, opened 2022-04-11
fix minor collection name first name

codiebeulaine:


rudigiesler (approved):
👍 Looks good. Should we have a test for it?

PR #455, opened 2022-04-12
Momza 2073 dev of opt out flow

Buhle79:


erikh360 (commented):
I think first sort out the errors and you don't have to commit the gruntfile changes so you can revert those.

I also think we don't need a v2 file here, as this is github and we have proper version control and more control over deploys compared to rapidpro.

PR #456, opened 2022-04-13
ussd new optout

Buhle79:


src/ussd_popi_rapidpro.js
```diff
@@ -927,37 +927,195 @@ go.app = function() {
...
+                    " " +
+                    "Have a lovely day!"
+                )
+            });
+        });
+
+        self.add("state_anonymous_data_optout", function(name, opts){
+            var msisdn = utils.normalize_msisdn(self.im.user.addr, "ZA");
+            var answers = self.im.user.answers;
+            var opt_out = answers.state_stop_being_part === "yes";
```
DevChima: You might want to use _.toUpper() here and  "YES" instead

src/ussd_popi_rapidpro.js
```diff
@@ -927,37 +927,195 @@ go.app = function() {
...
+
+        self.add("state_stop_being_part_optout_success", function(name) {
+            return new EndState(name, {
+                next: "state_start",
+                text: $(
+                    "Your research consent has been withdrawn, and you have been removed from all research." +
+                    " " +
+                    "MomConnect will continue to send you helpful messages." +
+                    " " +
+                    "Goodbye."
```
DevChima, line 1018: This could be:
(
          "Your research consent has been withdrawn, and you have been removed from all research. " +
          "MomConnect will continue to send you helpful messages. Goodbye "
        ),        
OR 
You could use \n or join("\n")) if you want new lines.

go-app-ussd_popi_rapidpro.js
```diff
@@ -1076,37 +1076,195 @@ go.app = function() {
...
+                subscriptions.push("baby due on " + edd.format("DD/MM/YYYY"));
+            }
+
+            if (dobs.length > 0) {
+                console.log('dobs.length > 0');
+                if (dobs.length == 1) {
+                    subscriptions.push("baby born on " + dobs[0].format("DD/MM/YYYY"));
+                } else {
+                    dobs.forEach(function(dob, i) {
+                        subscriptions.push("baby born on " + dob.format("DD/MM/YYYY"));
```
DevChima, line 1215: Can you please test this postbirth option too. "baby born on.."
EDD looks good.

DevChima: I see it's the entire block is commented out. We'll need to test both

src/ussd_popi_rapidpro.js
```diff
@@ -927,37 +927,195 @@ go.app = function() {
...
         });
 
-        self.add("state_opt_out", function(name) {
+        self.states.add("state_optout_menu", function(name) {
             return new MenuState(name, {
-                question: $("Do you want to stop getting MomConnect messages?"),
+                question: $("What would you like to do?"),
+                choices: [
+                    new Choice("state_opt_out_reason", $("Stop getting messages")),
+                    new Choice("state_stop_being_part", $("Stop being part of research")),
```
DevChima: state_stop_research might be easier to read than state_stop_being_part beyond this screen.

src/ussd_popi_rapidpro.js
```diff
@@ -993,76 +1153,33 @@ go.app = function() {
...
         });
 
         self.add("state_submit_opt_out", function(name, opts) {
             var msisdn = utils.normalize_msisdn(self.im.user.addr, "ZA");
             var answers = self.im.user.answers;
-            var forget = answers.state_delete_data === "state_delete_confirm";
-            var loss = answers.state_loss_messages === "state_loss_delete_data";
-            var loss_forget = answers.state_loss_delete_data === "yes";
+            var loss = answers.state_loss_messages === "state_submit_opt_out";
+            var loss_forget = answers.state_submit_opt_out === "yes";
```
DevChima: You can use _.toUpper() here and "YES"

test/ussd_popi_rapidpro.test.js
```diff
@@ -2205,163 +2119,49 @@ describe("ussd_popi_rapidpro app", function() {
...
-                        "2. Back"
-                    ].join("\n")
+                        "Sorry we don't understand. Please try again.",
+                        "1. Stop getting all MomConnect messages"
+                    ].join("\n"),
                 })
                 .run();
         });
-    });
+    });*/
```
DevChima: Trailing slash

go-app-ussd_popi_rapidpro.js
```diff
@@ -1076,37 +1076,195 @@ go.app = function() {
...
             return new MenuState(name, {
-                question: $("Do you want to stop getting MomConnect messages?"),
+                question: $("What would you like to do?"),
+                choices: [
+                    new Choice("state_opt_out_reason", $("Stop getting messages")),
+                    new Choice("state_stop_being_part", $("Stop being part of research")),
+                    new Choice("state_anonymous_data", $("Make my data anonymous")),
+                    new Choice("state_no_optout", $("Nothing. I still want to get messages"))
+                ],
+                error: $("Try again, don't understand."),
```
DevChima: We could use a more comprehensible error message here.

go-app-ussd_popi_rapidpro.js
```diff
@@ -1076,37 +1076,195 @@ go.app = function() {
...
 
-        self.add("state_no_optout", function(name) {
+        self.add("state_anonymous_data", function(name) {
             return new MenuState(name, {
-                question: $(
-                    "Thanks! MomConnect will continue to send helpful messages and process " +
-                    "your personal info. What would you like to do?"
+                question: $("If you make your data anonymous, we'll delete your phone number, " +
+                            "and we won't be able to send you messages." ,
+                            " " ,
```
DevChima: I don't think we need this "". You could use \n of join(\n)

go-app-ussd_popi_rapidpro.js
```diff
@@ -1076,37 +1076,195 @@ go.app = function() {
...
+                ]
+            });
+        });
+
+        self.add("state_no_optout", function(name) {
+            return new EndState(name, {
+                next: "state_start",
+                text: $(
+                    "Thanks! MomConnect will continue to send you helpful messages." +
+                    " " +
```
DevChima: ""

DevChima:
Looking great. I only left a few comments that are mostly cosmetic.

src/ussd_popi_rapidpro.js
```diff
@@ -1532,105 +1651,6 @@ go.app = function() {
             });
         });
 
-        self.add("state_all_questions_view", function(name) {
```
erikh360: Why are these states being removed?

src/ussd_popi_rapidpro.js
```diff
@@ -993,76 +1153,33 @@ go.app = function() {
...
-                ),
-                error: $(
-                    "Sorry we don't recognise that reply. Please enter the number next to your " +
-                    "answer."
-                ),
-                choices: [
-                    new Choice("state_submit_opt_out", $("Next"))
-                ]
+        self.add("state_message_unhelpful_or_unknown", function(name) {
+            return new EndState(name, {
```
erikh360: Shouldn't this also go to state_submit_opt_out?

src/ussd_popi_rapidpro.js
```diff
@@ -927,37 +927,195 @@ go.app = function() {
             });
         });
 
-        self.add("state_opt_out", function(name) {
+        self.states.add("state_optout_menu", function(name) {
```
erikh360: I think this should just be self.add

src/ussd_popi_rapidpro.js
```diff
@@ -927,37 +927,195 @@ go.app = function() {
...
+                question: $("If you make your data anonymous, we'll delete your phone number, " +
+                            "and we won't be able to send you messages." ,
+                            " " ,
+                            "Do you want to continue?"),
+                error: $(
+                    "Sorry we don't recognise that reply. Please enter the number next to your " +
+                    "answer."
                 ),
+                choices: [
+                    new Choice("state_anonymous_data_optout", $("Yes")),
```
erikh360, line 969: As per the new comments in Miro, this should also ask for a reason. I'm happy if we change this in a new PR to keep the size of this one down

src/ussd_popi_rapidpro.js
```diff
@@ -927,37 +927,195 @@ go.app = function() {
...
+
+        self.add("state_anonymous_data_optout", function(name, opts){
+            var msisdn = utils.normalize_msisdn(self.im.user.addr, "ZA");
+            var answers = self.im.user.answers;
+            var opt_out = answers.state_stop_being_part === "yes";
+            return self.rapidpro
+                .start_flow(
+                    self.im.config.optout_flow_id, null, "whatsapp:" + _.trim(msisdn, "+"), {
+                        opt_out: opt_out ? "TRUE" : "FALSE",
+                        mqr_consent: "Denied",
```
erikh360: Does the flow expect this field?

erikh360 (commented):
Looking good but a few comments

src/ussd_popi_rapidpro.js
Buhle79: @erikh360 I've removed them as we don't have them on the miro flow. Must i add them back ?

go-app-ussd_popi_rapidpro.js
Buhle79: @DevChima The block we are skipping for now i will add the test when we switching back to it.

src/ussd_popi_rapidpro.js
Buhle79: Will add to new PR

src/ussd_popi_rapidpro.js
Buhle79: Will add to new PR

src/ussd_popi_rapidpro.js
```diff
@@ -927,37 +927,194 @@ go.app = function() {
             });
         });
 
-        self.add("state_opt_out", function(name) {
+        self.add("state_optout_menu", function(name) {
```
erikh360, line 930: I don't think we need to rename this state, the mainmenu still references "state_opt_out"

erikh360: oh wait, I see what's happening, I think we just need to change the reference in "state_main_menu" to point here

src/ussd_popi_rapidpro.js
```diff
@@ -975,8 +1132,10 @@ go.app = function() {
                 next: function(choice) {
                     if(_.includes(["miscarriage", "stillbirth", "babyloss"], choice.value)) {
                         return "state_loss_messages";
+                    } else if(_.includes(["not_useful", "unknown"], choice.value)) {
+                        return "state_message_unhelpful_or_unknown";
```
erikh360: I think this should still go to state_submit_opt_out for these reasons. maybe split after that to show the specific messages.

src/ussd_popi_rapidpro.js
```diff
@@ -993,76 +1152,33 @@ go.app = function() {
...
+                        "\n" +
+                        "For any medical concerns, please visit a clinic."
+                )
             });
         });
 
         self.add("state_submit_opt_out", function(name, opts) {
             var msisdn = utils.normalize_msisdn(self.im.user.addr, "ZA");
             var answers = self.im.user.answers;
-            var forget = answers.state_delete_data === "state_delete_confirm";
```
erikh360, line 1057: I think we can leave this in since the forget should also submit through here.

erikh360: Ideally we shouldn't change this payload at all, this functionality isn't changing, just the way we present the questions

src/ussd_popi_rapidpro.js
```diff
@@ -927,37 +927,194 @@ go.app = function() {
...
+                    }
+                    return self.states.create(name, opts);
+                });
+        });
+
+        self.add("state_stop_research_optout", function(name, opts){
+            var msisdn = utils.normalize_msisdn(self.im.user.addr, "ZA");
+            return self.rapidpro
+                .start_flow(
+                    self.im.config.clear_pii_flow, null, "whatsapp:" + _.trim(msisdn, "+")
```
erikh360: Instead of `clear_pii_flow` lets use something like `research_optout_flow`, we can create a new flow called `External Opt Out - Research` for this in rapidpro where we just change mqr_consent to denied and research_consent to FALSE

erikh360 (commented):
A few more comments

src/ussd_popi_rapidpro.js
erikh360: this comment is not fixed yet?

src/ussd_popi_rapidpro.js
```diff
@@ -993,77 +1142,38 @@ go.app = function() {
...
 
         self.add("state_submit_opt_out", function(name, opts) {
             var msisdn = utils.normalize_msisdn(self.im.user.addr, "ZA");
             var answers = self.im.user.answers;
-            var forget = answers.state_delete_data === "state_delete_confirm";
-            var loss = answers.state_loss_messages === "state_loss_delete_data";
-            var loss_forget = answers.state_loss_delete_data === "yes";
+            var forget = _.toUpper(answers.state_anonymous_data) === "YES";
+            var loss = _.toUpper(answers.state_loss_messages) === "YES";
+            var loss_forget = _.toUpper(answers.state_submit_opt_out) === "YES";
```
erikh360: These are choicestates, the value won't be YES, that is only the label that the user sees, see the lines above you replaced

src/ussd_popi_rapidpro.js
erikh360: Looks like this one is also not resolved

src/ussd_popi_rapidpro.js
Buhle79: Okay i see

src/ussd_popi_rapidpro.js
erikh360: I se this in the forget part that we can do in the next PR

erikh360 (approved):

PR #457, opened 2022-04-19
Update optout line with new changes

erikh360:
Sorry about the size

### DIFF
--- a/go-app-ussd_optout_rapidpro.js
+++ b/go-app-ussd_optout_rapidpro.js
@@ -167,8 +167,8 @@ go.app = function() {
                     // Delegate to the correct state depending on group membership
                     var contact = self.im.user.get_answer("contact");
                     if(self.contact_in_group(contact, _.concat(self.im.config.public_group_ids, self.im.config.clinic_group_ids))){
-                        return self.states.create("state_opt_out");
-                    } else if(self.contact_in_group(contact, self.im.config.optout_group_ids && _.get(contact, "fields.optout_reason") === "")) {
+                        return self.states.create("state_optout_menu");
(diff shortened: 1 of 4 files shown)


DevChima:
I didn't compare with the miro board but code looks fine.

go-app-ussd_optout_rapidpro.js
```diff
@@ -334,32 +344,33 @@ go.app = function() {
...
             return self.rapidpro.start_flow(self.im.config.flow_uuid, null, "whatsapp:" + _.trim(msisdn, "+"), {
-                babyloss_subscription:
-                    self.im.user.get_answer("state_loss_optout") === "yes" ? "TRUE" : "FALSE",
-                delete_info_for_babyloss:
-                    self.im.user.get_answer("state_delete_info_for_loss") === "yes" ? "TRUE" : "FALSE",
-                delete_info_consent:
-                        self.im.user.get_answer("state_delete_research_info") === "yes" ? "TRUE" : "FALSE",
+                babyloss_subscription: loss,
+                delete_info_for_babyloss: forget ? "TRUE" : "FALSE",
+                delete_info_consent: forget ? "TRUE" : "FALSE",
```
rudigiesler: Having `forget` be boolean and `loss` be text can be a bit confusing. If we make them the same (either both TRUE/FALSE or both boolean), then I think that removes that confusion, and can probably avoid making a mistake with the type.

rudigiesler (approved):

go-app-ussd_optout_rapidpro.js
erikh360: good point, thanks

PR #459, opened 2022-05-26
tbcheck ussd update

Buhle79:


src/ussd_tb_check.js
```diff
@@ -660,7 +660,12 @@ go.app = function () {
       return new JsonApi(self.im)
         .post(self.im.config.healthcheck.url + "/v2/tbcheck/", payload)
         .then(
-          function () {
+          function (response) {
+            self.im.user.answers = {
```
erikh360: Won't his overwrite the whole answers object with a new one? I think it's better to set each value separately

src/ussd_tb_check.js
```diff
@@ -691,40 +695,226 @@ go.app = function () {
         question: text,
         error: error,
         accept_labels: true,
-        choices: [new Choice("state_show_results", $("See Results"))],
+        choices: [new Choice("state_display_arm_message", $("See Results"))],
       });
     });
 
-    self.states.add("state_show_results", function (name) {
```
erikh360, line 698: Shouldn't we still show the original message if study is not active yet and arm is empty?

src/ussd_tb_check.js
```diff
@@ -691,40 +695,226 @@ go.app = function () {
...
+    self.states.add("state_display_arm_message", function (name) {
       var answers = self.im.user.answers;
-      var risk = self.calculate_risk();
-      var text = $(
-        "You don't need a TB test now, but if you develop cough, fever, weight loss " +
-          "or night sweats visit your nearest clinic."
-      );
+      var arm = answers.group_arm;
+      return self.states.create("state_" + arm);
```
erikh360: This can be one line if we only use the arm in the one place.

erikh360 (commented):
Looks good so far, just a few comments.

Please also make sure you have tests around all the new states to see that the flow is correct.

src/ussd_tb_check.js
Buhle79: @erikh360 Which one is the original message ?

src/ussd_tb_check.js
Buhle79: Oh yes it does override it thanks.

src/ussd_tb_check.js
```diff
@@ -660,7 +660,14 @@ go.app = function () {
       return new JsonApi(self.im)
         .post(self.im.config.healthcheck.url + "/v2/tbcheck/", payload)
         .then(
-          function () {
+          function (response) {
+            answers.group_arm = response.data.tbconnect_group_arm.toLowerCase();
+            answers.tbcheck_id = response.data.id;
+            /*self.im.user.answers = {
```
erikh360: Can we remove this completely please.

erikh360 (approved):
just the one small comment

PR #464, opened 2022-07-14
skip address questions

Buhle79:


bearnard:
Looks good, before we merge we should test that the healthcheck app serializer validation works locally at least

bearnard:
👍 nice!

PR #465, opened 2022-07-19
changed validation for existing province

Buhle79:


test/ussd_tb_check.test.js
```diff
@@ -801,6 +802,14 @@ describe("ussd_tb_check app", function () {
         })
         .run();
     });
+    it("should skip the state for users who already have province", function () {
```
bearnard: add and are from activation `skip_location_2022`

bearnard:
👍

PR #473, opened 2022-08-18
research_consent should be true not "YES"

DevChima:


bearnard:
there seem to be other places that look for a string eg: https://github.com/praekeltfoundation/ndoh-jsbox/blob/develop/src/ussd_tb_check.js#L274

Since the django app accepts boolean values you need to be sure jsbox sends `false` when they refuse consent

DevChima:
@bearnard If they reply "No" they exit the app and we don't do a SEND. https://github.com/praekeltfoundation/ndoh-jsbox/blob/1d2df8c19e6012c00bda244f101af6cf1fdf0274/src/ussd_tb_check.js#L295

bearnard:
ah, cool, does `ChoiceState` not handle this for us?

bearnard:
👍 otherwise

PR #474, opened 2022-08-18
Revert "research_consent should be true not "YES""

DevChima:
Reverts praekeltfoundation/ndoh-jsbox#473


DevChima:
For some reason the docker image did not build so reverted.

PR #494, opened 2022-08-29
Add USSD source to pass serializer

DevChima:
https://github.com/praekeltfoundation/healthcheck/blob/4574d692284e140b5be494d42e322691cb5ac30e/tbconnect/serializers.py#L21


Buhle79:
+1

PR #510, opened 2022-09-28
synced missing translation

Buhle79:


Buhle79:
Please not we still have outstanding translations

bearnard:
👍 When will they be ready? We generally want to keep the develop branch deployable/clean

PR #523, opened 2022-11-21
Not skip street and suburb for users who want to change it

DevChima:
At state_confirm_address if a user replies "No", they can only change the city but no the suburb or street because we only check if it exists and then skip if TRUE.


Buhle79 (approved):

bearnard:
👍

PR #524, opened 2022-11-21
Suburb and street fix

DevChima:
{.....,"state_confirm_city":"state_street_name"},"metadata":{},"in_session":true,"state":{"name":"state_city","metadata":{},"creator_opts":{}}}

What's saved the menustate is the literal name which in this case is the state name not the displayed string (No or Yes).


Buhle79 (commented):
+1

PR #525, opened 2022-11-21
updated tracing screen for the study

Buhle79:


go-app-ussd_tb_check.js
```diff
@@ -886,6 +886,14 @@ go.app = function () {
     });
 
     self.add("state_exposure", function (name) {
+      var next_state = "state_tracing";
+//      var activation = self.get_activation();
```
bearnard: why the comment out?

Buhle79: Checking

bearnard:
👍

PR #543, opened 2023-02-16
Tbh 552 tb check afrikaans copy changes sun feedback

Buhle79:


config/go-app-ussd_tb_check.afr_ZA.json
```diff
@@ -130,7 +130,7 @@
     ],
     "FAQ for more info on TBCheck and the research.": [
         null,
-        "FAQ, vir meer oor TBCheck asook die navorsing daaroor."
+        "FAQ vir meer information."
```
erikh360: I think this should be "informasie"?

Buhle79: @erikh360 

I copied from their feedback and pasted as is

PR #554, opened 2023-04-17
Fix date formats

erikh360:
... and added some logging

### DIFF
--- a/go-app-ussd_mqr_faqs.js
+++ b/go-app-ussd_mqr_faqs.js
@@ -241,19 +241,30 @@ go.app = function() {
         });
 
         self.is_content_stale = function() {
+            console.log(">>> is_content_stale");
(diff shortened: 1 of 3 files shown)


go-app-ussd_mqr_faqs.js
```diff
@@ -241,19 +241,30 @@ go.app = function() {
         });
 
         self.is_content_stale = function() {
+            console.log(">>> is_content_stale");
```
Buhle79, line 244: Did you leave these print statements on purpose?

Buhle79 (commented):
+1

go-app-ussd_mqr_faqs.js
erikh360: Yes, will clean up at a later stage, thanks for the review!

PR #583, opened 2023-08-01
Move pending msisdn change trigger + add confirm

erikh360:
There won't be an active subscription on the new msisdn until after its confirmed so we need to move the check to the no active subscription branch.


go-app-ussd_popi_rapidpro.js
```diff
@@ -301,6 +301,18 @@ go.app = function() {
                 });
         });
 
+        self.states.add("state_msisdn_change_completed", function(name) {
```
erikh360, line 304: Adding this state to give the flow a chance to do the actual switch, then back to state start to get the updated contact.

Buhle79 (commented):
+1

PR #587, opened 2023-08-02
Use prebirth messaging field to check for prebirth instead of edd field

DevChima:
- Use prebirth messaging field to check for prebirth instead of the EDD date field.
     This is necesary because we do not clear EDD when we do a baby switch from prebirth to postbirth.

- Also the EDD siwtch confirmation message shows the new EDD as Datetime where as we only want to see the date.

### DIFF
--- a/go-app-ussd_popi_rapidpro.js
+++ b/go-app-ussd_popi_rapidpro.js
@@ -44,6 +44,53 @@ go.Hub = function() {
     return Hub;
 }();
 
+go.Engage = function() {
+    var vumigo = require('vumigo_v02');
+    var events = vumigo.events;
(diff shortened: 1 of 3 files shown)


go-app-ussd_popi_rapidpro.js
```diff
@@ -44,6 +44,53 @@ go.Hub = function() {
     return Hub;
 }();
 
+go.Engage = function() {
```
erikh360: We shouldn't have this block in here at all, maybe do a pull to update your gruntfile

erikh360 (commented):
+1 if you fix the comment

PR #595, opened 2023-09-05
Add new_wa_id to msisdn change

DevChima:
Contacts who go through the nosim_msisdn_change route would not have the new_wa_id sent to Rapidpro.
It wasn't part of the payload even though the RP flow expects it. 

This pr adds it to the list so it gets sent to RP.


erikh360 (approved):
Very nice! 🍰

PR #596, opened 2023-09-05
Add sms disabled flag clinic reg

erikh360:
Looks big but its just the new rapidpro function duplicated in all the go-* files.


DevChima:
Cool. This is the endpoint that lists globals right

erikh360:
@DevChima yes, so we only have to change the config in one place to disable

DevChima:
> @DevChima yes, so we only have to change the config in one place to disable

Super!

PR #608, opened 2023-10-16
Tbh 710 bug missing screen after language selection

Buhle79:


DevChima:
After adding tests, if it pass then 👍

PR #626, opened 2023-11-07
add test to check that research consent is undefined

Hlamallama:


erikh360 (commented):
I think we'll also want to test that the payload after the check is correct because isn't that where the code changed?

Buhle79 (approved):

PR #643, opened 2024-08-15
Fix EDD and DoB display for change info

Buhle79:


go-app-ussd_popi_rapidpro.js
```diff
@@ -478,14 +478,28 @@ go.app = function() {
...
+             else {
+                context = {
+                    dobs: _.map(_.filter([
+                        new moment(baby_dob1),
+                        new moment(baby_dob2),
+                        new moment(baby_dob3),
+                    ], _.method("isValid")), _.method("format", "DD-MM-YYYY")).join(", ") || $("None")
+                 };
+             }
+
```
DevChima, line 502: @Buhle79 I now remember why we didn't consider a (dob but no edd) option because a contact with dob would have an edd and at the time we weren't clearing the edd field in RP (not sure if we now do). But it's good to have this check.

DevChima (approved):
Looks good. I was going to ask that you add one more test with edd, dob1, dob2 and dob3 but I see it's already there as an old test and it still works.

PR #646, opened 2024-10-28
users can now register a baby on day of birth

Mudiwa66:


Mudiwa66:
I had to change some test cases

erikh360 (approved):

PR #651, opened 2025-07-16
changed clinic ussd from rapidpro to turn

Mudiwa66:
The Turn.js file was added to connect with the turn api. The src/ussd_clinic_rapidpro.js was updated to connect to Turn.

### DIFF
--- a/src/ussd_clinic_rapidpro.js
+++ b/src/ussd_clinic_rapidpro.js
@@ -19,14 +19,14 @@ go.app = function() {
         var $ = self.$;
 
         self.init = function() {
-            self.rapidpro = new go.RapidPro(
+            self.Turn = new go.Turn(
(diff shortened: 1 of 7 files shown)


src/ussd_clinic_rapidpro.js
```diff
@@ -19,14 +19,14 @@ go.app = function() {
         var $ = self.$;
 
         self.init = function() {
-            self.rapidpro = new go.RapidPro(
+            self.Turn = new go.Turn(
```
erikh360: Please make a copy of this file called ussd_clinic_turn.js and only apply the changes to the new file.

ussd_clinic_rapidpro.js should be left unchanged so that we can still maintain existing USSD lines while we work on this. Thanks

